PR #21, opened 2015-01-30
Remove dependency on pomegranate

expez:
This is pretty ugly, but at least it works!

If we're going down this route we should try to get pomegranate patched so `add-dependencies` will take an optional classloader.  Most of the code here is because we cannot call that function directly.


benedekfazekas:
looks like black magic, but after today's discussion actually makes sense. nice work @expez!

expez:
hmm, the `resource-paths` isn't being overridden by the `dev` profile so the tests fail.  Any idea how to fix that?

benedekfazekas:
have not tested but don't really get how this works without the transient dependencies of pomegranate... what do I miss?

benedekfazekas:
got error when testing manually:
in the `*Messages*` buffer:
```
Added enlive version 1.1.5 as a project dependency
error in process filter: progn: Failed to load ClojureRuntimeImpl
error in process filter: Failed to load ClojureRuntimeImpl
```

nothing in the nrepl-server buffer. new dependency not hotloaded

expez:
hmm, maybe this only worked because I already have everything pomgranate needs on classpath (or even loaded due to my `profiles.clj`?)

benedekfazekas:
tests do pass. but I guess we have only tests for the non pomegranate part of add dependency

expez:
ya, this only worked because I had pomegranate available on the classpath.  After clearing out `profiles.clj` I get the same error as you.

I changed the `build` section of the pomegranate `pom.xl` to 

```xml
  <build>
    <resources>
      <resource>
        <directory>src/main/clojure</directory>
      </resource>
    </resources>

    <plugins>
      <plugin>
        <artifactId>maven-assembly-plugin</artifactId>
        <configuration>
          <archive>
            <manifest>
              <mainClass>fully.qualified.MainClass</mainClass>
            </manifest>
          </archive>
          <descriptorRefs>
            <descriptorRef>jar-with-dependencies</descriptorRef>
          </descriptorRefs>
        </configuration>
      </plugin>
    </plugins>
```

and built a jar containing all transitive deps with `mvn clean compile assembly:single` (this jar also contains clojure).  When I load it into shimdandy and attemp to require pomegranate I just get the same error.

benedekfazekas:
well, this part:
```
              (URL. "file:libs/shimdandy-impl-1.1.0.jar")
              (URL. "file:libs/clojure-1.6.0.jar")
              (URL. "file:libs/pomegranate-0.3.0.jar")
```
won't work either. because there won't be a `libs` directory when you use refactor-nrepl in a project. and it does not make much sense to package jars inside a jar I think. unzipping them and adding them to the refactor-nrepl jar in a distinct path and then create a separate classloader for them, shimdandy etc may work tho

expez:
I think jars in jars are OK, as the classloader works on jars.You're right that the plugin is failing because the paths are wrong when it's all packaged into a jar, but in the repl the paths are right but I can't get the `rt` to require pomegranate.

expez:
I've pushed another commit containing the huge pomegranate jar

benedekfazekas:
you mean in the repl for refactor-nrepl itself the paths are right? this may be. but it won't work with any other project as you can't package jars into a jar.

src/refactor_nrepl/externs.clj
```diff
@@ -0,0 +1,28 @@
...
+  (let [urls [(URL. "file:libs/shimdandy-impl-1.1.0.jar")
+              (URL. "file:libs/clojure-1.6.0.jar")
+              (URL. "file:libs/pomegranate-0.3.1-SNAPSHOT.jar")]
+        cl (URLClassLoader. (into-array URL urls))]
+    (reset! rt (ClojureRuntimeShim/newRuntime cl "refactor-nrepl"))))
+
+(defn add-dependencies [dep repos]
+  (when-not @rt
+    (create-runtime))
+  (.require @rt (into-array String ["cemerick.pomegranate"]))
```
expez: @tobias when testing this code at the repl this require fails when it cannot find `pomegranate` on the classpath.  Any ideas as to why? 

 Furthermore, if I remove all the jars except for `shimdandy-impl` I can still call `(.invoke @rt "clojure.core/+" 1 2)` and get an answer.  I'd expect that to fail.  Is it working because the new runtime has the old classloader as a parent?

tobias: The classloader cannot find pomegranate is because there is no pomegranate jar in lib/ in the current directory. If you are putting the pomegranate jar inside the refactor jar, a URLClassLoader cannot see it when given a file: url, since it's not a file. Unfortunately, you can't give the URLClassLoader a jar: url pointing at the jar within the jar, since it does not support that. If you want to ship jars within jars and then add the inner jars to the classpath, you have to extract them to disk first, then use file: urls that point to the files on disk.

You are correct, Shimdandy can still see the clojure runtime because the parent classloader can see it. Shimdandy doesn't change the way classloader delegation works, so if clojure.jar is visible to a higher classloader, that version gets used. You can't properly use Shimdandy from within a clojure application without writing a custom classloader - it's intended to be used from a java application that then manages multiple runtimes.

For what you are trying to do, it may be easier to use https://github.com/flatland/classlojure, or just create a URLClassLoader, point it to a pomegranate jar somewhere on disk, and call `(.setContextClassLoader (Thread/currentThread) your-cl)` before calling anything that requires pomegranate. You'll probably want to set the TCCL back to what it was originally when you are done.

expez: @tobias I was testing this at the repl (hoping to solve the jar-in-jar problem later, @benedekfazekas  suggested just unpacking everything into a dir in the jar and pointing the classloader in that direction) but I still couldn't get the pomegranate code to eval properly. 

I think it finds the lib folder and the jars correctly.  Since the shim runtime is created it must find, `shimdandy-impl` right?  

Inheriting from existing classloaders is OK, I think.  If I understand how it works then the nearest classloader is searched first and then if something isn't found the parent is searched.  As long as we're not polluting the user's classpath or classloader we should be golden.

Thanks for taking a look!

tobias: Ah, good point - it does have to see the lib/ dir for shimdandy to create a runtime. I'm not sure why it can't load pomegranate then. Can you share a gist of the stack you get?

Normal delegation has the classloader ask it's parent *first*, then search itself if the parent doesn't have it (see http://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html). So the parent's version of clojure will always be used. This also means that any namespaces required in the shimdandy runtime will be added to the internal lib cache in the "global" clojure runtime, but the classes would not be visible to that runtime when shimdandy's classloader isn't active. I'm not sure what sort of problems that would cause.

expez:
```
  Show: Clojure Java REPL Tooling Duplicates All  (22 frames hidden)

1. Unhandled java.io.FileNotFoundException
   Could not locate cemerick/pomegranate__init.class or
   cemerick/pomegranate.clj on classpath:

                       RT.java:  443  clojure.lang.RT/load
                       RT.java:  411  clojure.lang.RT/load
                      core.clj: 5530  clojure.core/load/fn
                      core.clj: 5529  clojure.core/load
                   RestFn.java:  408  clojure.lang.RestFn/invoke
                      core.clj: 5336  clojure.core/load-one
                      core.clj: 5375  clojure.core/load-lib/fn
                      core.clj: 5374  clojure.core/load-lib
                   RestFn.java:  142  clojure.lang.RestFn/applyTo
                      core.clj:  619  clojure.core/apply
                      core.clj: 5413  clojure.core/load-libs
                   RestFn.java:  137  clojure.lang.RestFn/applyTo
                      core.clj:  619  clojure.core/apply
                      core.clj: 5496  clojure.core/require
                   RestFn.java:  408  clojure.lang.RestFn/invoke
                      Var.java:  415  clojure.lang.Var/invoke
   ClojureRuntimeShimImpl.java:   57  org.projectodd.shimdandy.impl.ClojureRuntimeShimImpl/require
 NativeMethodAccessorImpl.java:   -2  sun.reflect.NativeMethodAccessorImpl/invoke0
 NativeMethodAccessorImpl.java:   62  sun.reflect.NativeMethodAccessorImpl/invoke
DelegatingMethodAccessorImpl.java:   43  sun.reflect.DelegatingMethodAccessorImpl/invoke
                   Method.java:  483  java.lang.reflect.Method/invoke
                Reflector.java:   93  clojure.lang.Reflector/invokeMatchingMethod
                Reflector.java:   28  clojure.lang.Reflector/invokeInstanceMethod
                          REPL:    1  refactor-nrepl.externs/eval27247
```

tobias: I'm not sure what's going on there. Can you give me a quick HOWTO on how I can recreate it here?

expez: @tobias just clone the repo and check out the `shimdandy` branch. Then run following in the repl, after evaluating `externs.clj`:

```clj 
(create-runtime)
(.require @rt (into-array String ["cemerick.pomegranate"])
```

expez: @tobias  did you manage to reproduce this, or do you need some more details?

tobias: I was able to recreate it, and in the process of debugging, I removed shimdandy to reduce the moving parts. With shimdandy out of the way, I was able to come up with a version that works just using clojure internals: https://gist.github.com/tobias/f2cbb026977299c8c71e

Let me know if that works for you.

expez:
Classlojure looks like the easiest to use

PR #34, opened 2015-03-05
Fix resolve-missing for records and types

expez:
This also removes an assumption that capitalized names belong to
classes.  This prevented the resolution of e.g. schemas from
prismatic/schema, which are vars that conventionally start with capital
letters.

The return value of the op has changed.  We're now dropping :type because
it no longer contains the resolution we need on the client side.  We have
to be able determine the difference between a missing:

1) record or type, as created by defrecord or deftype, respectively
2) class
3) ns

The new return value is thus an alist of ((candidate1 . type1) (candidate
2 . type2) ...), where type is in #{:type :ns :class}.

A :type is generated by both defrecord or deftype becaues the client will
do the same in both cases.

Interfaces are included under :class because info-clj doesn't make the
distinction and the client doesn't have to branch on the information.

It's quite convenient to simply read in result alist on the elisp
side, but I imagine this might not be the case in clients written in other
languages.  If we ever get other consumers it should be easy enough to
change the representation, but I don't want to write that code now so I'm
hoping for YAGNI :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The goal of the op is to provide intelligent suggestions when the user wants to 
 
 The op requires `symbol` which represents a name to look up on the classpath.
 
-The return value `candidates` is a space separated string of candidates to import or require into the current ns.
+The return value `candidates` is an alist of `((candidate1 . type1) (candidate2 . type2) ...)` where type is in `#{:type :class :ns}` so we can branch on the 3 various way to import.  `:type` means the symbol resolved to a var created by `defrecord` or `deftype`, `:class` also includes interfaces.
 
 ### hotload-dependency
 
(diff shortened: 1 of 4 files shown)


expez:
This functionality will break unless you update `clj-refactor` while grabbing the snapshot of `refactor-nrepl` I'm about to publish

expez:
snapshot 18 of 0.3 is out.

rundis:
Nice !

rundis:
@expez 
Weird. It seems to me that the latest snapshot on clojars doesn't contain this fix. (nuked my `~/.m2/repository/refactor-nrepl/refactor-nrepl/0.3.0-SNAPSHOT` forced download, unzipped and checked that the change was not included in for example resolve_missing.clj)

expez:
I just put this into `install.sh` to make sure I remembered to use `mranderson`:

```sh
#!/bin/bash
lein do clean, source-deps
lein with-profile +plugin.mranderson/config deploy clojars
```

...and then I ran it on the wrong branch.

I've uploaded snapshot 19, and now you know why the count is so high :grinning:

rundis:
:-)

And voila I have some additional cool features to add ! tx.

Issue #18, opened 2015-01-21
refactor-nrepl interferes with existing working web app


hellofunk:
I thought you might like to see this issue:

https://github.com/dakrone/clj-http/issues/249

I had this in my lein plugins:

`:plugins [[refactor-nrepl "0.3.0-SNAPSHOT"]]`

As soon as I removed it, the Java error disappeared and my app started working again. I was not actually using the refactor tools when the crash happens.

I'd like to keep using the refactor functionality, so if you have advice on what might have cause this, I'm all ears.

bbatsov:
Most likely - a dependency conflict.

hellofunk:
Resolving the conflict is proving to be no small chore.

benedekfazekas:
yeah I think so. I think it is related to pomegranate which depends on wagon-http which depends on an older version of `org.apache.httpcomponents` than `clj-http`

we use [mranderson](https://github.com/benedekfazekas/mranderson) to minimise these conflicts. but this guy is basically a bunch of java classes: mranderson serves no remedies for those...

I think we had a chat with @expez at some point to try to get rid of pomegranate for this reason: but not easy if we want to have resolve artifacts...

benedekfazekas:
you can try to explicitly define the higher httpcomponents version in your project's project.clj and see if that solves the problem...

benedekfazekas:
alternatively we could manually inline httpcomponents (and wagon http, pomegranate?!) as java source. sounds horrible but doable... @expez what do you think?

hellofunk:
I am definitely not going to reorganize my own working project code just so I can use refactor tools in an editor.

benedekfazekas:
fair point @hellofunk

expez:
@benedekfazekas I don't see any other option than inlining

benedekfazekas:
we might get away with inlining just a few java classes which are really used by hotload... don't know. really guessing around here...

expez:
are [boot's pods](https://github.com/boot-clj/boot/wiki/Pods) the solution to this and all similar problems?  If I'm reading this right we could put `refactor-nrepl` in its own pod and do away with mranderson?

bbatsov:
I think they might be! I'm thinking of using them for cider-nrepl itself.

On 22 January 2015 at 15:47, Lars Andersen <notifications@github.com> wrote:

> are boot's pods <https://github.com/boot-clj/boot/wiki/Pods> the solution
[...]

PR #36, opened 2015-03-06
Upgrade Apache httpclient dependency to 4.4

nwolfe:
Similar to #18...

I just spent quite a bit of time figuring out why [`org.apache.commons.codec.DigestUtils#getString(String)`](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/DigestUtils.html#getDigest(java.lang.String)) wasn't a public method during my SSL library unit tests (which has a dependency on `commons-codec "1.9"`).

Turns out a newer version of `httpclient` resolves this issue for me. 
* I bumped the version #, `lein install`ed it, started my Emacs Cider REPL, and did some refactorings - things seemed to still work.
* `lein test :all` passes locally

Also, let me give my :+1: to resolving these dependency issues as mentioned in #18

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,7 +10,7 @@
                  ^:source-dep [http-kit "2.1.19"]
                  [org.clojure/data.json "0.2.5"]
                  ^:source-dep [instaparse "1.3.4"]
-                 ^:source-dep [org.apache.httpcomponents/httpclient "4.3.5"]
+                 ^:source-dep [org.apache.httpcomponents/httpclient "4.4"]
                  ^:source-dep [com.cemerick/pomegranate "0.3.0"]
                  ^:source-dep [org.clojure/tools.analyzer.jvm "0.6.5"]
                  ^:source-dep [org.clojure/tools.namespace "0.2.7"]


nwolfe:
*sigh* Guess this wasn't enough...getting new exceptions now.

It looks like I can downgrade to `refactor-nrepl "0.2.2"` without getting my original `DigestUtils` problem, but now I get exceptions when trying to perform some of the refactorings (like rename symbol).

Guess I'm going to have to live with a broken refactor-nrepl for now.  Closing this PR since the newer version only breaks things in a different (but seemingly less-frequent) way.

benedekfazekas:
oi @nwolfe do u still have a problem with 0.3 snapshot?

nwolfe:
@benedekfazekas Yeah, sorry should've mentioned the version I was using when I filed this.  This is broken for me in 0.3.0-SNAPSHOT.

benedekfazekas:
ok. what u need to override is common codecs. i will push it tomorrow. but the root of the problem is our dependency on pomegranate pulling all this stuff in.  we should really get rid of that

nwolfe:
> what u need to override is common codecs. i will push it tomorrow.

Ah yeah, that would work. Thank you - I'll look for it!

benedekfazekas:
solved in latest 0.3.0-SNAPSHOT

PR #40, opened 2015-03-07
gets rid of cider as a provided dependency

benedekfazekas:
- inlines `info-clj` and `java/resolve-symbol` from cider

sort of think that this still has value as we don't 'force' users (both end users and other possible editor integrations) to use cider middleware too. not 100% sure tho hence the pull request. cc @expez, @bbatsov @rundis

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,8 +16,7 @@
                  ^:source-dep [org.clojure/tools.reader "0.8.12"]]
   :plugins [[thomasa/mranderson "0.3.1"]]
   :filespecs [{:type :bytes :path "refactor-nrepl/refactor-nrepl/project.clj" :bytes ~(slurp "project.clj")}]
-  :profiles {:provided {:dependencies [[cider/cider-nrepl "0.8.2"]]}
-             :test {:dependencies [[print-foo "1.0.1"]]}
+  :profiles {:test {:dependencies [[print-foo "1.0.1"]]}
              :1.5 {:dependencies [[org.clojure/clojure "1.5.1"]]}
              :1.6 {:dependencies [[org.clojure/clojure "1.6.0"]]}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0-alpha5"]]}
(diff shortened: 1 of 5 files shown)


bbatsov:
This is the wrong line of reasoning IMO. I've aways hoped to merge portions of refactor-nrepl into cider-nrepl when the deps issues are cleared. Seems strange to separate the basic functionality from the refactoring functionality - normally one would build one on top of the other. I can only assume that most editors would benefit from having extra features at their disposal.

benedekfazekas:
I generally agree this is why I did not just push this and created the PR.

my only concern that there are use cases when ppl might want to use only refactor-nrepl but not cider. example: ppl who prefer inf-clojure still may be interested for certain functions from refactor-nrepl. also I recommended to @rundis to use the goodies from cider middleware too in his plugin but for his ligthtable plugin he does not strictly need the cider middleware.

expez:
I'm not sure I want to inline any more stuff.

I think we can leave this as `:provided` until we figure out how to isolate our stuff on the classpath for real.  I might have some time to fiddle with that next week.  

Other than working on classloader isolation we should be in polish mode and try to get `1.0` out the door.

benedekfazekas:
I am not itching to merge this. let's hear what @martintrojer and @rundis says if they want to share their opinion ;)

rundis:
tbh I haven't looked too much at cider-nrepl yet, but it seems to plenty of goodies that might be cool to add to the light-table plugin as well. My only concern would be if pulling in additional dependencies increases the chances of dependency conflicts with peoples projects.
At some point in the future my plugin (parts or the whole thing) might be merged into the existing [Clojure plugin](https://github.com/LightTable/Clojure) for Light Table.

Until that potentially happens  its no big deal if users of my plugin have to add 1 or 2 dependencies.

To summarize: Sounds to me the best option would be to leave it as a provided dep in your project and I'll live happily with stating that both needs to be a precondition for using my plugin.

PR #43, opened 2015-03-09
[Fix #29] Split refactor into several ops

### LINKED ISSUE #29: Split refactor.clj into 3 distinct ops
We now have a single op `refactor` in this namespace but then we branch on `refactor-fn` too.  It would be much better if we had 3 distinct ops.  This would make it possible to have a more meaningful `descriptor`, making `refactor-nrepl` more self-documenting, as well as making it possible to return more semantically meaningful return values than `value`, e.g. `:occurrences`.

expez:
- Rename refactor.clj => find_symbol.clj
- Create a new toplevel op find-symbol
- Create a new toplevel op find-debug-fns
- Change the name of some of the parameters to find-symbol:
  loc-line => line, loc-column => column, clj-dir => dir
- The return value of find-symbol is changed from [4 4 3 20 println ...]
  to a plist with the following keys: :line-beg :line-end :col-beg
  :col-end :name :path :match.
- Expanded the readme to better document find-symbol

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,31 @@ Takes one required argument, `artifact` which is the full name of the artifact e
 
 The return value is a space-separated list of all the available versions for the artifact.
 
-### find symbols
+### find-symbol
 
-Finds occurrences of symbols like defs and defns both where they are defined (if available) and where they are used.
+This op finds occurrences of a single symbol.
+
+`find-symbol` requires:
+
+`file` "The absolute path to the file containing the symbol to lookup.
+
+`dir` Only files below this dir will be searched.
+
+`ns` The ns where the symbol is defined.
+
+`name` The name of the symbol
+
+`line` The line number where the symbol occurrs.
+
+`column` The column number where the symbol occurs.
+
+The return value is a stream of occurrences under the key `occurrence` which is an alist like this:
+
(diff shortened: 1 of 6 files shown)


expez:
If this looks OK to you guys I'll make the corresponding changes in `clj-refactor`.

@benedekfazekas I haven't done anything about `find-debug-fns` because you indicated you were considering removing that op.  I've never used that functionality myself so I'm OK with removing that op.

Sorry about creating some more work for you @rundis.

benedekfazekas:
nice! only have time to have a look in the evening tho

benedekfazekas:
no objection to merge this. my only concern that users will need to upgrade both clj-refactor and the middleware. but better now then after 1.0.

expez:
new snapshot is out 

Breakage ahead @rundis

benedekfazekas:
:+1: I assume no problem with mranderson then...

expez:
It works wonderfully here, but I couldn't get a repl going on the windows machine at work earlier today

Hopefully it was a problem with the build itself so when I grab the snapshot tomorrow everything will be honkey dory

benedekfazekas:
I don't even have a windoz machine so I really hope I don't need to set a VM up or something for testing :P ;)

rundis:
Allrightie then. I'm on it :)

PR #71, opened 2015-05-09
Add op stubs-for-interface

expez:
This returns enough information to create stub implementations of
protocols and interfaces in the client.

Specifically it returns the name of functions to implement as well as
the parameter list.

https://github.com/clojure-emacs/clj-refactor.el/pull/148


expez:
Great, tests are passing on some versions of clojure but not others.  :(

PR #73, opened 2015-05-17
Tidy up

luxbock:
I hope this sort of stuff is welcome. I had some trouble understanding what was going on so I decided to refactor things around a little bit and to make everything a bit more readable.

I tried to make the commits small in case some of the changes are not welcome.

### DIFF
--- a/src/refactor_nrepl/find_symbol.clj
+++ b/src/refactor_nrepl/find_symbol.clj
@@ -47,9 +47,6 @@
   (let [fn-set (into #{} (str/split fn-names #","))]
     (find-nodes ast (partial fns-invoked? fn-set (util/alias-info ast)))))
 
-(defn- contains-var [var-name alias-info node]
-  (contains-var? #{var-name} alias-info node))
-
 (def ^:private symbol-regex #"[\w\.:\*\+\-_!\?]+")
 
 (defn- contains-const?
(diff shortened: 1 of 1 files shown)


benedekfazekas:
making the code better/more readable deffo wlc. thanks for making small separate commits, you are right we might cherry pick a bit around here.

again: :+1:

luxbock:
I just realized after re-starting my REPL that the tests aren't passing anymore. I'll try to figure out what went wrong.

benedekfazekas:
> I just realized after re-starting my REPL that the tests aren't passing anymore. I'll try to figure out what went wrong.

thx

benedekfazekas:
did a quick run through. the ones i did not comment need a bit more thought/code checking from my part. also waiting for the build to be fixed.

ta!

luxbock:
The tests failing appear to have been an issue with my `profiles.clj` conflicting with the project, or at least now that I removed my customizations, running `lein with-profile +plugin.mranderson/config test` passes everything.

benedekfazekas:
failing on travis tho i think

--
Benedek Fazekas

benedekfazekas:
https://travis-ci.org/clojure-emacs/refactor-nrepl/builds/62922129

luxbock:
I'm really having a hard time understanding what's wrong. I used `git bisect` to figure out which commit caused the problem: 3f0599f. There was a mistake, which I fixed and rebased it using a `fixup` commit. Then I ran `build.sh` again, but the tests were still failing. I couldn't figure out what else it could be, so I created a new branch from the previous commit that still succeeded, and applied a commit with exactly the same changes. Running `build.sh` on the new branch passes all the tests. This doesn't make any sense to me, as its content should be exactly the same as the commit that was failing before.

PR #78, opened 2015-05-26
Allow :format "none" to be specified by clients

clumsyjedi:
so that client software which cannot interpret clojure can still
interact with the bencode data

### DIFF
--- a/src/refactor_nrepl/middleware.clj
+++ b/src/refactor_nrepl/middleware.clj
@@ -35,6 +35,13 @@
   `(with-errors-being-passed-on ~transport ~msg
      (transport/send ~transport (response-for ~msg ~(apply hash-map kv)))))
 
+(defn serialize-response [{:keys [serialization-format] :as msg} response]
+  (condp = serialization-format
+    "edn" (pr-str response)
+    "bencode" response
+    (pr-str response) ; edn as default
+    ))
+
 (defn resolve-missing-reply [{:keys [transport] :as msg}]
   (reply transport msg :candidates (resolve-missing msg) :status :done))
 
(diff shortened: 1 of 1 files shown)


src/refactor_nrepl/middleware.clj
```diff
@@ -35,6 +35,11 @@
   `(with-errors-being-passed-on ~transport ~msg
      (transport/send ~transport (response-for ~msg ~(apply hash-map kv)))))
 
+(defn format-response [{:keys [format] :as msg} response]
```
expez: I think a better name here would be `serialize-response` or something along those lines.  To me `format-response` makes me think about pretty printing more than producing different data formats.

src/refactor_nrepl/middleware.clj
```diff
@@ -35,6 +35,11 @@
...
+  (if (= format "none")
```
expez: I think we should switch this around and do `(= format "edn")` so people aren't left wondering what exactly `none` means.

bbatsov: Yep, "none" is pretty confusing IMO.

expez:
There's also a capitalization mistake in the commit message that you might fix while you're addressing the other remarks.

I think other people might be interested in this, so adding a note about this in the `README` might not be such a bad idea.  If you do that, you should also write a test to ensure we don't break this.

I'm OK with leaving this as an easteregg, though, since any bugs will be filed over in your repo :)

src/refactor_nrepl/middleware.clj
```diff
@@ -35,6 +35,13 @@
   `(with-errors-being-passed-on ~transport ~msg
      (transport/send ~transport (response-for ~msg ~(apply hash-map kv)))))
 
+(defn serialize-response [{:keys [serialize-format] :as msg} response]
```
expez: This should probably be `serialization-format`

expez:
Fix the key name and squash the commits and we should be good to go.

I'm fine with no tests / documentation while you figure out if this is something that's going to give you any value at all.

clumsyjedi:
@expez cool, thanks for the feedback. Key has been renamed, commits have been squashed.

expez:
Great, thanks! :+1:

PR #80, opened 2015-06-02
Add rename-file-or-dir

expez:
This op is used to move a file or a directory of files.  If any clojure
files are affected by the move any dependent namespaces are updated.

Unlikely tools.namespace's move-ns this op also handles vectors written
in prefix notation as well as fully qualified references in the file
body itself referencing vars from the old namespace.

The return value of the op is a list of the file which were affected by
the move.  This list does not include any deleted files. These files all
have to be visited by the client to indent the changed ns form until we
get proper pretty printing support in the middleware.

This ops is quite slow because it does multiple passes over each file
but it's still orders of magnitude faster than doing this manually so
I'm not that worried about speed in this initial commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,18 @@ This op returns, `version`, which is the current version of this project.
 
 This eagerly builds, and caches, ASTs for all clojure files in the project.  Returns `status` `done` on success.
 
+### rename-file-or-dir
+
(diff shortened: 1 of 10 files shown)


benedekfazekas:
:+1:

PR #92, opened 2015-07-03
Update file path of profile.boot in README.md

xhh:
according to boot's wiki page: https://github.com/boot-clj/boot/wiki/Configuring-Boot#configuring-your-project


expez:
Thanks! :+1:

PR #96, opened 2015-07-08
Rewrite clean-ns

expez:
The previous implementation relied on the AST to to find some symbols
and a manual walk to find the rest.  Due to the effect of macros we're
already walking so much of the file that we might as well abandon the
AST and walk everything. The new implementation thus walks every form
in the body and collects symbols to use when pruning libspecs and
imports.

The benefit of this approach are as follows

- We're for sure not going to remove a libspec or a referred symbol
  which is needed.
- Since we don't rely on an AST the op will work on files that are in a
  'bad state'.
- This approach should be a lot easier to adapt to .cljs and .cljc files.

The disadvantage of this approach is that we're going to leave some
symbols and libspecs around in the ns form when they are unused.  This
should be limited to symbols found in the :refer clause, and the
community seems to be moving away from referring symbols in favor of
aliasing.

A future commit will take locals into account to limit the amount of
false positives we get when we're looking for a symbol.  This will
improve the accuracy if this op, but I don't think it'll ever get
perfect.

This fixes clojure-emacs/clj-refactor.el#186

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,7 +16,8 @@
   :plugins [[thomasa/mranderson "0.4.0"]]
   :filespecs [{:type :bytes :path "refactor-nrepl/refactor-nrepl/project.clj" :bytes ~(slurp "project.clj")}]
   :profiles {:provided {:dependencies [[cider/cider-nrepl "0.9.0"]]}
-             :test {:dependencies [[print-foo "1.0.1"]]}
+             :test {:dependencies [[print-foo "1.0.1"]]
+                    :src-paths ["test/resources"]}
              :1.5 {:dependencies [[org.clojure/clojure "1.5.1"]]}
              :1.6 {:dependencies [[org.clojure/clojure "1.6.0"]]}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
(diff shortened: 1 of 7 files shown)


bbatsov:
Hmm, why not put this in `cider-nrepl` then? Without the AST dep it seems like an ideal candidate to the promoted to cider itself. The users will love this.

expez:
> Hmm, why not put this in cider-nrepl then?

It's not done yet and I don't want to maintain it in two places.

> Without the AST dep it seems like an ideal candidate to the promoted to cider itself. 

This still depends on instaparse, but it doesn't have to.  When I first wrote `clean-ns` I was really excited about how awesome instaparse is, so I had a bad case of "hammer looking for nail" and ended up using instaparse when I should've done without.  

I wrote a little instaparse grammar to parse the ns form, instead of doing this by hand, but this didn't really result in less code and it's going to be a problem when adding cljs because the grammar would have to take into account reader macros everywhere and anywhere and at that  point it won't be easy to work with at all.

bbatsov:
> This still depends on instaparse, but it doesn't have to. When I first wrote clean-ns I was really excited about how awesome instaparse is, so I had a bad case of "hammer looking for nail" and ended up using instaparse when I should've done without.
> 
> I wrote a little instaparse grammar to parse the ns form, instead of doing this by hand, but this didn't really result in less code and it's going to be a problem when adding cljs because the grammar would have to take into account reader macros everywhere and anywhere and at that point it won't be easy to work with at all.

Sounds familiar. :-) I'm looking forward to seeing this simplified.

PR #109, opened 2015-08-20
Add cljs to clean-ns

expez:
There's now support for:

* pruning of libspecs and imports
* sorting of libspecs and imports
* removal of :use
* :refer-macros are pruned and sorted
* :require-macros are pruned and sorted
* :use-macros are re-written to :require-macros

For now I'm not going to update the readme or  the changelog because this
is likely to still be somewhat rough around the edges.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ pom.xml.asc
 install.cmd
 /install.sh
 /deploy.sh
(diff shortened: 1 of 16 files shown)


benedekfazekas:
wohoo awesome!

PR #113, opened 2015-08-30
Make find-macros aware of cljc files, require 1.7

expez:
Macros defined in cljc files are now correctly found.  Their usage in
clj files is correctly tracked.

In order to deal with cljc files we're going to require clojure 1.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: clojure
 script:
   - lein2 source-deps :prefix-exclusions "[\"classlojure\"]"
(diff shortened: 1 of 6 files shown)


expez:
Dropping support for anything below 1.7 going forward.

`tools.namespace` is going to require 1.7 from now on (it's written using reader macros now), so when we want to add more cljs goodness that'll force us to upgrade anyway.

//cc @magnars @benedekfazekas @bbatsov @Malabarba

PR #114, opened 2015-09-05
Add cljs support to resolve-missing

expez:
For now this only handles macros and regular cljs vars.  Adding support
google closure classes will be tackled later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ install.cmd
 /.cljs_nashorn_repl/
 /.cljs_rhino_repl/
 /nashorn_code_cache/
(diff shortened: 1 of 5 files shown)


expez:
What do you guys think we should do in a cljc file where we can't determine the context?  Should we prompt the user for the language context?  Should we return all candidates regardless of language?  If so, how do present the candidates so the user can spot the difference between vars with the same fully qualified name?

Malabarba:
Maybe do one at a time?
Prompt the user for the clj require, then for the cljs require, and add both to the `ns` form.

PR #116, opened 2015-09-21
Teach rename-file-or-dir how to handle cljs

expez:
This also handles the case where we have a macro defined in clojure
which is used in cljs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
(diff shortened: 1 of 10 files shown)


expez:
I'm pretty sure you can now rename `cljc` files too, as long as no other `cljc` files are affected.

benedekfazekas:
if you mean the mranderson case i guess it `cljc` files all the way down ;). just look into `tools.namespace "0.3.0-alpha1"`

PR #117, opened 2015-09-24
Teach clean-ns how to handle cljc files

expez:
This works but it's not as pretty as I'd like.

There are no pretty printing facilities for cljc files, and I'm not
motivated to write any, so the ns is printed using
`clojure.pprint/pprint`.

The simplest thing was to wrap all the dependency forms of each language in
a reader macro.  This makes editing, and reading, very easy, but the ns
form now contains far more duplication than is strictly necessary.

The option to do `prefix-rewriting` is now ignored in cljc files.  I
felt this was a good idea because cljs doesn't have the prefix notation
and it would look weird (and complicate matters) if we didn't preserve
symmetry.

The ns looks like this:

```
(ns my-namespace
   #?@
   (:clj
    [(:require
       [clojure.walk :as walk]
       .
       .
       .
                              ])
    :cljs
    [(:require [cljs.test :refer-macros [is deftest]]
    .
    .
    .
                             ])))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New features
 
+* Add cljc support to `clean-ns`.
+* Add cljs support to `clean-ns`.
 * `rename-file-or-dir` now knows how to move cljs files.
 * The `artifact-version` list is now sorted.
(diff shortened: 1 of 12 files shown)


expez:
Manually applied to master

PR #120, opened 2015-09-28
Fix find symbol in refer regression

benedekfazekas:
- renames `ns.namespace-aliases` to `ns.libspecs`
- uses indexing pushback reader when parsing the ns from so the location
  info is recorded/retained
- gets rid of the temp directory for integration tests

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -3,6 +3,7 @@
             [clojure.java.io :as io]
             [clojure.string :as str]
             [clojure.tools.namespace.parse :as parse]
+            [clojure.tools.reader.reader-types :as readers]
             [me.raynes.fs :as fs]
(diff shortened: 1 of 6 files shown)


test/resources/testproject/src/com/example/five.clj
```diff
@@ -57,7 +57,3 @@
   (let [{:keys [foo bar] :or {foo "foo"}} (hash-map)]
     [:bar :foo]
     (count foo)))
-
-;; This was causing both find-local-symbol and find-macros to blow up, for
-;; different reasons
-::str/bar
```
benedekfazekas: no idea what this is but it breaks the project for me (tests don't even start running)

expez: I added it because symbols like that have caused errors in two of our refactorings before.  I don't want a section in the readme about unsupported clojure constructs so I don't want to merge this until this problem is solved.

PR #125, opened 2015-10-18
More relaxed with invalid code

benedekfazekas:
Default to ignore errors for `find symbol`. This results `find usages`, `rename symbol` and `inline symbol` to work even if there is a broken ns in the project.

Also exploit some `tools.analzyer` AST validation hooks so namespaces with type errors, unresolved symbols or possible arity problems will be still analyzed.

`clj-refactor.el` (and other clients) will need a `cljr-block-on-analyzer-errors` so strictness loving users can block above features on AST problems.

this PR is basically a result of a discussion on 6th Oct between @expez @Malabarba and myself on gitter


benedekfazekas:
as discussed on gitter. will merge tmrw if no considerations raised

benedekfazekas:
merged manually

PR #144, opened 2016-02-16
Fix resolve-missing for cljs

magnars:
Resolve-missing returns a list of candidates. Before this fix, the
cljs-path returned only a single candidate. That's because every
candidate was merged together. We now use `merge-with into` to generate
a list of candidates instead.


expez:
Awesome! Thanks :+1:

PR #147, opened 2016-02-21
Avoid needless .cljc namespace reader macro usage

### LINKED ISSUE #143: Needless .cljc namespace reader macro usage
I have a .cljc file with this ns:

```clojure
(ns example.foo
  (:require [example.bar :as bar]))
```

clean-ns turns this into:

```clojure
(ns example.foo
  #?@(:clj
       [(:require [example.bar :as bar])]
       :cljs
       [(:require [example.bar :as bar])]))
```

This seems needlessly verbose, as the form is the same in both cases.

magnars:
Fixes #143


expez:
Yes! :+1:

PR #151, opened 2016-03-06
Add support for skipping excluded dependencies

arrdem:
This patch tweaks the dependency checker, so that refactor-nrepl will activate itself normally if one of its dependencies is listed in a project's :exclusions. This enables a user to for instance exclude
org.clojure/clojure and provide a fork, thus taking on all responsibility for providing an adequate version.


expez:
Thanks @arrdem :)

arrdem:
I'm moderately surprised that you just merged that. Thanks Lars!

expez:
Well, it'll probably only benefit you, but it's a small patch with no real downsides so it's fine :)

PR #157, opened 2016-04-07
Config setting for exceptional libspecs

expez:
Some libspecs are side-effecting at load time but appear to be unused
otherwise.  `clean-ns` will need a list of these to ignore.  As such
namespaces might be user-created, and I don't want the responsibility of
maintaining a curated list of every such occurrence, this has to be
user-editable.

### DIFF
--- a/src/refactor_nrepl/ns/prune_dependencies.clj
+++ b/src/refactor_nrepl/ns/prune_dependencies.clj
@@ -3,7 +3,8 @@
             [refactor-nrepl
              [core :as core]
              [util :as util]]
-            [refactor-nrepl.find.symbols-in-file :as symbols-in-file]))
+            [refactor-nrepl.find.symbols-in-file :as symbols-in-file]
+            [refactor-nrepl.config :as config]))
 
 (defn- lookup-symbol-ns
   ([ns symbol]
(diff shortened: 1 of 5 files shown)


src/refactor_nrepl/ns/prune_dependencies.clj
```diff
@@ -110,10 +111,15 @@
...
+;; pruned.
 (defn- libspec-should-never-be-pruned? [libspec]
-  ;; The symbols of cljsjs namespaces are externs, and unknowable for now,
-  ;; so we just ignore these libspecs
-  (.startsWith (str (:ns libspec)) "cljsjs"))
+  (let [ns-name (str (:ns libspec))]
+    (reduce (fn [acc ^String pattern]
+              (or acc (re-find (re-pattern pattern) ns-name)))
+            false
+            (:ignore-libspecs config/*config*))))
```
benedekfazekas: I would rather use `some` here

```
(defn- libspec-should-never-be-pruned? [libspec]
  (let [ns-name (str (:ns libspec))]
    (some
     (fn [^String pattern]
       (re-find (re-pattern pattern) ns-name))
     (:ignore-libspecs config/*config*))))
```

expez: Agreed :+1:

PR #161, opened 2016-04-25
Add dirty namespace tracking for resolve-missing

### LINKED ISSUE #160: resolve-missing doesn't find newly defined symbols due to cache
I'm using `resolve-missing` a lot from `clj-refactor.el` to add missing require/refer statements to the namespace declaration.

I noticed that when defining new functions and using them from another namespace I had to require them manually, `refactor-nrepl` would not find them. It would work fine after restarting the Clojure process.

I finally traced it down to the fact that in `refactor-nrepl.ns.slam.hound.regrow` the symbols are cached, so anything defined/loaded later on gets ignored.

I worked around this by disabling the cache, and now it works great.

``` emacs-lisp
(add-hook 'nrepl-connected-hook
          (lambda ()
            (cider-nrepl-request:eval
             "(alter-var-root #'refactor-nrepl.ns.slam.hound.regrow/*cache* (constantly nil))"
             (lambda (x)))))
```

I would like to help to come up with a general solution, but I'm not sure what the best course of action is. Dropping the caching is one option, but I assume it's there for a reason. Alternatively it would need some kind of invalidation, maybe marking namespaces as "dirty" when evaluating forms in them.

Thoughts?

plexus:
resolve-missing internally caches a symbol->namespace mapping. New vars that
were added later are not in the cache, and so can not be found.

This code maintains a list of namespaces that were manipulated through nREPL,
and will recheck these every time, incrementally adding to the cache.

Currently this list of dirty namespaces is append only. It will grow over time,
but should still remain small compared to the whole of all loaded namespaces.

This represents a half way point. This same tracking can be extended for other
cached functions in slam.hound.regrow, such as for imports. When that happens
removing namespaces from the dirty list can only be done when all caches have
been updated or invalidated.

Closes #160 

This is my first PR here, it took some searching around. I'd appreciate any feedback.

### DIFF
--- a/src/refactor_nrepl/ns/slam/hound/regrow.clj
+++ b/src/refactor_nrepl/ns/slam/hound/regrow.clj
@@ -10,6 +10,14 @@
            (java.util.regex Pattern)))
 
 (def ^:dynamic *cache* (atom {}))
+(def ^:dynamic *dirty-ns* (atom #{}))
+
+(defn wrap-clojure-repl [f]
+  (fn [& args]
+    (swap! *dirty-ns* conj *ns*)
+    (apply f args)))
+
+(alter-var-root #'clojure.main/repl wrap-clojure-repl)
 
 (defmacro ^:private caching [key & body]
   `(if *cache*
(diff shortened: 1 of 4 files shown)


src/refactor_nrepl/ns/slam/hound/regrow.clj
```diff
@@ -20,6 +28,22 @@
          v#))
      (do ~@body)))
 
+(defmacro cache-with-dirty-tracking
+  "The function to be cached, f, should have two signatures. A zero-operand
+  signature which computes the result for all namespaces, and a two-operand
+  version which takes the previously computed result and a list of dirty
+  namespaces, and amends the result."
```
expez: You probably didn't mean 'amends' here.

expez:
> This same tracking can be extended for other cached functions in slam.hound.regrow, such as for imports. 

This means new stuff defined with `defrecord` or `deftype` won't be found, right?  If that's the case then we should probably open an issue here and fix that too.  

Overall I think this looks good.  The wrapping of `clojure.main/repl` looks a bit gross at first glance, but I think this is OK.  What we're doing should compose with whatever anyone else is doing.

We're missing an update to the changelog and I'd love it if you prefixed the commit message with `[Fix #160]` so we get a nice and uniform git history.

src/refactor_nrepl/ns/slam/hound/regrow.clj
plexus: does it sound too much like mutating? "updates"?

expez: I've just never heard anyone talk about 'amending' a datastructure :) I think 'updates' is better, yeah.

plexus:
I added dirty tracking for imports as well. I left it in a separate commit for now so you can see what's changed.

`all-ns-imports*` is a bit of a mess at the moment, I'm going to come back to it with fresh eyes. I'll squash and update the changelog and the commit message then as well.

expez:
Awesome! 👍

plexus:
Still need to squash and fix the commit message, but code-wise I think I'm done.
 
In the previous version I tried to filter out classes from java.lang because these are included in every namespace. This was misguided. The function is five times faster without that filtering.

When hotloading dependencies this cache needs to be reset. This was already happening for java classes, but not for clojure symbols and types. It's a bit tricky to write a test for that, but I tried it on a project, `cljr-add-project-dependency`, then `cljr-add-missing-libspec` and it finds the newly loaded identifiers which is :100:. It does mean that first call after hotloading a new dependency is slow (cold cache). That's the price you pay.

expez:
> It does mean that first call after hotloading a new dependency is slow (cold cache). That's the price you pay.

We could update the cache in the background, in another thread.  I've also been thinking about adding an op to warm this cache.  The first call at work takes like 20s.  

We can get to that later.  This is still a huge improvement! :+1:

plexus:
* [x] squashed and cleaed up comit message
* [x] updated CHANGELOG.md
* [x] ran ./build.sh

I think it's done now :) :arrow_right: :green_apple:

expez:
Wonderful work @plexus!  Thanks a lot! :+1:

expez:
A snapshot with this fix is now on clojars

PR #165, opened 2016-06-25
Replace (reduce concat ,,,) with (mapcat ,,,)

plexus:
The (reduce concat ,,,) pattern was causing the stack to overflow, this
was a bit tricky to debug, the symptom was that CIDER would just block
forever.

Also factored out a use of `proxy` so the class generation only needs to
happen once. All in all this version seems to perform significantly
better than the previous one.

(That last point I can't exactly prove since the old version was no longer working for me, so I would appreciate if someone could compare the performance of both versions.  `(get-available-classes)` now runs in <20ms where I seem to remember it being more in 500~1000ms range.)

### DIFF
--- a/src/refactor_nrepl/ns/slam/hound/search.clj
+++ b/src/refactor_nrepl/ns/slam/hound/search.clj
@@ -34,15 +34,17 @@
 
 ;; could probably be simplified
 
+(def jar-filter
+  (proxy [FilenameFilter] []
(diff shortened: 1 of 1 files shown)


expez:
This code is copy-paste from slamhound which is a pretty old project.  I kind of assumed the code to be near bug-free, so didn't read through it that closely.  Thanks for digging into this! :+1:

PR #166, opened 2016-07-14
Fix a NullPointerException in nrepl-refactor.core/prefix

plexus:
When a file contains the symbol `'/` (e.g. `(/ 6 2)`), then `symbols-in-file`
through `fix-ns-of-backquoted-symbols` will try to find the prefix of that
symbol. The prefix function thinks the slash is a namespace separator, and
things blow up.

This seems to have been introduced in

3839db7 Fix clean-ns with namespaced keywords

Before `(prefix '/)` would simply return `nil`, this fix restores the old behavior.


expez:
Thanks!

PR #170, opened 2016-09-16
Update README.md

caioaao:
Added newest versions for `refactor-nrepl` and `cider-nrepl`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ The role of this nREPL middleware is to provide refactoring support for clients 
 Add the following, either in your project's `project.clj`,  or in the `:user` profile found at `~/.lein/profiles.clj`:
 
 ```clojure
-:plugins [[refactor-nrepl "2.2.0"]
-          [cider/cider-nrepl "0.11.0"]]
+:plugins [[refactor-nrepl "2.3.0"]
+          [cider/cider-nrepl "0.14.0"]]
 ```
 
 ### Adding the middleware via Boot


magnars:
`2.3.0` of `refactor-nrepl` is not yet released.

benedekfazekas:
neither cider-nrepl 0.14.x. This section is a bit outdated tho in the README thanks for drawing my attention to it @caioaao

caioaao:
The problem is that emacs modes are already on versions 0.15.0 and 2.3.0 and show error messages when using different versions.

expez:
Could you paste the warning?

2.3.0 clj-refactor isn't out yet either, but it's true that cider now has a new snapshot out.

caioaao:
Using versions 0.14.0 and 2.2.0 on repl and doing `cider-connect`:
```
WARNING: CIDER's version (0.15.0-snapshot) does not match cider-nrepl's version (0.14.0). Things will break!
         More information.
Their versions are 2.3.0-SNAPSHOT (package: 20161005.344) and 2.2.0, respectively.
You can mute this warning by changing cljr-suppress-middleware-warnings.
```
Package versions: 
- cider: `20161016.424` (installed as dependency to `clj-refactor`)
- clj-refactor: `20161005.344`

benedekfazekas:
the readme refers to the latest stable versions. 2.3.0 of clj-refactor is not yet out. you get the warnings because your lein or boot profiles refer to the stable versions of the middlewares while you using the latest emacs clients.

you either need to sync those versions (use snapshot versions of the middlewares or use stable versions of the clients). you also won't have this problem if you use `cider-jack-in` instead of `cider-connect` if your context allows that.

caioaao:
ok, my bad. Just realized I'm using `melpa` instead of `melpa-stable` for a long time

benedekfazekas:
no worries :)

PR #178, opened 2016-12-15
[#Fix 167] Only build AST for files on classpath

expez:
Does this look like an acceptable fix, @benedekfazekas?


benedekfazekas:
give me till this evening to have a look

expez:
Thanks! :+1:

PR #184, opened 2017-01-20
Remove laziness in Slamhound's "get-available-classes"

plexus:
Once the classpath gets big enough, this would cause a stack overflow, which
from clj-refactor simply shows up as a timeout.

Instead use transducers over lists as much as possible. Since the collection of
available classes is only ever traversed sequentially, this provides the best
performance as well.

- [x] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x] You've added tests (if possible) to cover your change(s) --> functionality hasn't changed, so no tests added
- [x] All tests are passing (run `lein do clean, test`)
- [x] Code inlining with mranderson works and tests pass with inlined code (run `./build.sh install` -- takes a long time)
- [x] You've updated the changelog (if adding/changing user-visible functionality) --> no user visible changes
- [x] You've updated the readme (if adding/changing user-visible functionality) --> no user visible changes

Thanks!

PR #185, opened 2017-01-22
Report problems of type java.lang.Error back to the nREPL client

plexus:
Error inherits directly from Throwable, so Errors are not Exceptions, they need
to be handled separately. This category includes RuntimeError,
StackOverflowError, and NoClassDefFoundError.

In regular applications you would not catch these, but in our case letting them
bubble means the nREPL handler dies without returning a response, causing the
client to assume that the operation is taking too long and time out, without any
useful feedback.

This is bad UX, and it makes it hard to diagnose and debug problems like #184.
At least this way the user will see a stack trace that hints at the problem.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 *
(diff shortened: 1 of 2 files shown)


expez:
Sweet, thanks! :+1:

PR #186, opened 2017-01-22
Handle NoClassDefFoundError when determining class type

plexus:
When `resolve-missing` gets a number of candidate classes, it will load them to
determine the type (Clojure type/record vs regular Java class). However a class
on the classpath may be compiled with dependencies that are not currently
present, causing the loading to fail.

This would cause the whole operation to fail, instead this change handles the
exception, takes it as a sign that it's a regular Java class, and continues.

For example, I do a `resolve-missing` on the symbol URL, this yields two
candidates: `java.net.URL`, and
`com.gargoylesoftware.htmlunit.javascript.host.URL`. The latter depends on
`org.w3c.dom.ElementTraversal`, which is not currently present, causing a
`NoClassDefFoundError`.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 *
+* [#186](https://github.com/clojure-emacs/refactor-nrepl/issues/186) Make sure `resolve-missing` still works, even if a candidate class has missing dependencies.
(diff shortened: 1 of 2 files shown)


expez:
Merged this one manually after trying the 'resolve conflicts' button on github and getting an unexpected result.

Thanks for the patch! :+1:

PR #187, opened 2017-01-23
Upgrade clojure.tools.reader to support namespaced map literals

plexus:
And one more :)

`clean-ns` was failing on namespaces that contain Clojure 1.9 style namespaced maps. Because the included version of tools reader was pretty far out of date. This bumps the version from 0.10.0-alpha3 to 1.0.0-beta4. Namespaced maps are supported since tools.reader 1.0.0-beta3.

Is there a general policy on when to bump dependencies? There are quite a few that could be updated. Also, would it make sense to add a `:1.9` profile to test for Clojure 1.9 compatibility?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed
 
+* [#187](https://github.com/clojure-emacs/refactor-nrepl/issues/187) Update clojure.tools.reader so Clojure 1.9 namespaced map literals don't cause parse errors.
 * [#185](https://github.com/clojure-emacs/refactor-nrepl/issues/185) Report throwables of type `Error` instead of swallowing them.
 * [#186](https://github.com/clojure-emacs/refactor-nrepl/issues/186) Make sure `resolve-missing` still works, even if a candidate class has missing dependencies.
 * [#184](https://github.com/clojure-emacs/refactor-nrepl/pull/184) In `resolve-missing`, prevent classpaths with many entries from causing a stack overflow.
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,7 @@
                  ^:source-dep [alembic "0.3.2"]
                  ^:source-dep [org.clojure/tools.analyzer.jvm "0.6.9"]
                  ^:source-dep [org.clojure/tools.namespace  "0.3.0-alpha3"]
-                 ^:source-dep [org.clojure/tools.reader "0.10.0-alpha3"]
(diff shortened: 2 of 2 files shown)


expez:
> Is there a general policy on when to bump dependencies? There are quite a few that could be updated. 

The only goal we have is to be backward compatible as far back as is reasonably possible.  So far this has meant back to 1.7.  

> Also, would it make sense to add a :1.9 profile to test for Clojure 1.9 compatibility?

For sure

benedekfazekas:
LGTM

benedekfazekas:
> Also, would it make sense to add a :1.9 profile to test for Clojure 1.9 compatibility?

+1 also would be nice to add this to the test matrix in the build file. however the build already takes like 10 mins we should probably look into it and make it a bit snappier (long build is fine on travis but a bit annoying when developing, checking PRs etc)

expez:
Thanks! :+1:

magnars:
Oh, very nice! I need this. :) Any chance for a new snapshot release?

By the way, I tried `lein with-profile plugin.mranderson/config install`, but I ended up getting:

```
Their versions are 2.3.0-SNAPSHOT and n/a, respectively.
```

And no refactor middleware.

expez:
@magnars I got that after upgrading CIDER at home, but it seemed to work anyway.  We rely on cider's code to display these warnings...

I'll push a snapshot later today if @benedekfazekas doesn't beat me to it.

magnars:
@expez I hoped that would be the case as well, but got this warning when trying to use it: `user-error: The refactor-nrepl middleware isn't available! Did you remember to install it?`

benedekfazekas:
will try to beat @expez ;)

but meanwhile try `./build.sh install` @magnars

benedekfazekas:
or else just run
`lein do clean, source-deps :prefix-exclusions "[\"classlojure\"]"`
before running
`lein with-profile plugin.mranderson/config install`

(`./build.sh install` should work too but running tests is kinda sloooow)

benedekfazekas:
completely forgot about this. sry. shall i upload a snapshot version or already done and all parties are fine?
/cc @magnars @expez

expez:
Go for it @benedekfazekas :)

plexus:
thanks, that'd be great. Maybe a "real" release would also be nice at some point. Some of these fixed bugs are really annoying when you happen to run into them.

benedekfazekas:
snapshot release: on it.

we are one issue away from 2.3.0 afaik

benedekfazekas:
new 2.3.0-snapshot is on clojars

PR #189, opened 2017-01-31
clean-ns: Introduce ^:keep metadata on libspecs

plexus:
This allows marking up required namespaces that should not be cleaned by clean-ns.

This is useful to require namespaces only so they get loaded for side-effects. It can also be used in `(ns user)` to make functions available for the REPL.

Example

``` clojure
    (ns user
      (:require [com.stuartsierra.component :as component]
                ^:keep [reloaded.repl :refer [system init start stop go reset]]])
```

This PR is not ready yet, I'm just at the point where I got it to work, and I'd appreciate feedback.

Questions:

- Is `^:keep` the best key to use? (bikeshedding time!)
- Right now it's only supported in `:require`, does it make sense to also have it for `:import` or `:use`. (Probably not since you don't usually `:import` for side effects, and `:use` is deprecated, but maybe there's an argument to be made).
- Should it also work on individuals function names in a `:refer` macro?
- Should it work on nested vectors when grouping by prefix?

Todo for me:

- [ ] Check if it also works on CLJS/CLJC
- [ ] Update CHANGELOG
- [ ] Update README

----

- [ ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)

Thanks!

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -245,12 +245,16 @@
   metadata which shouldn't be printed back out."
   [file-content]
   (let [ns-string (sexp/get-first-sexp file-content)
-        meta? (-> ns-string (.replaceFirst "\\^\\{" "\\{")
-                  (StringReader.)
-                  (PushbackReader.)
-                  parse/read-ns-decl
-                  second)
-        shorthand-meta?  (second (re-find #"\^:([^\s]+)\s" ns-string))]
+        ns-decl (-> ns-string
+                    (.replaceFirst "\\^\\{" "\\{")
+                    (StringReader.)
+                    (PushbackReader.)
+                    parse/read-ns-decl)
+        meta? (second ns-decl)
+        ns-name (if (symbol? meta?) meta? (first (nnext ns-decl)))
+        shorthand-meta? (-> (java.util.regex.Pattern/compile (str "\\^:([^\\s]+)\\s" ns-name))
+                            (re-find ns-string)
+                            second)]
     (cond
       (map? meta?) meta?
       shorthand-meta? {::shorthand-meta (keyword shorthand-meta?)}
--- a/src/refactor_nrepl/ns/ns_parser.clj
+++ b/src/refactor_nrepl/ns/ns_parser.clj
@@ -19,10 +19,11 @@
 
 (defn- libspec-vector->map
   [libspec]
-  (if (vector? libspec)
-    (let [[ns & specs] libspec]
-      (into {:ns ns} (->> specs (partition 2) (map vec))))
-    {:ns (symbol libspec)}))
+  (-> (if (vector? libspec)
+          (let [[ns & specs] libspec]
+            (into {:ns ns} (->> specs (partition 2) (map vec))))
+        {:ns (symbol libspec)})
+      (with-meta (meta libspec))))
 
 (defn- expand-prefix-specs
   "Eliminate prefix lists."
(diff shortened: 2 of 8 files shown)


src/refactor_nrepl/core.clj
```diff
@@ -245,12 +245,16 @@
...
-                  second)
-        shorthand-meta?  (second (re-find #"\^:([^\s]+)\s" ns-string))]
+        ns-decl (-> ns-string
+                    (.replaceFirst "\\^\\{" "\\{")
+                    (StringReader.)
+                    (PushbackReader.)
+                    parse/read-ns-decl)
+        meta? (second ns-decl)
+        ns-name (if (symbol? meta?) meta? (first (nnext ns-decl)))
+        shorthand-meta? (-> (java.util.regex.Pattern/compile (str "\\^:([^\\s]+)\\s" ns-name))
```
plexus: This code was too eager, it would transform

``` clojure
(ns user (:require ^:keep [,,,]))
```

into 


``` clojure
(ns ^:keep user)
```

expez:
> Is `^:keep` the best key to use? (bikeshedding time!)

I'd prefer `refactor-nrepl/keep`, that way it's possible for someone to find out what that metadata means even if they're not familiar with the tool.

> Right now it's only supported in `:require`, does it make sense to also have it for `:import` or `:use`. (Probably not since you don't usually `:import` for side effects, and `:use` is deprecated, but maybe there's an argument to be made).

Agree with you on this.  We can always revisit it later.

> Should it also work on individuals function names in a :refer macro?

I don't think that's needed, when you mark it as `keep` I think it's OK to interpret that as "I'll manage this one manually".  I might be biased because I almost never refer stuff.

> Should it work on nested vectors when grouping by prefix?

We do prefix re-writing today so we need at least partial support.

```clj
(ns my-ns 
  (:require 
    [foo.bar :as bar]
    ^:refactor-nrepl/keep 
    [foo.baz :as baz]))
```

With prefix rewriting this would look like:

```clj
(ns my-ns 
  (:require 
    [foo
      [bar :as bar]
      ^:refactor-nrepl/keep 
      [baz :as baz]]))
```

But I don't think we need to worry about:

```clj
(ns my-ns 
  (:require 
    ^:refactor-nrepl/keep 
    [foo
      [bar :as bar]
      [baz :as baz]]))
```

If it's easier to just support both, then that's fine I think, but I wouldn't bother adding complexity to support this final variant.

plexus:
> I'd prefer `:refactor-nrepl/keep`, that way it's possible for someone to find out what that metadata means even if they're not familiar with the tool.

That makes sense. I was thinking it could be a convention that other tools can also pick up, but that's probably not too likely anyway.

expez:
Ping @plexus :)

plexus:
Hey, sorry for letting this sit. I've been completely swamped with other stuff, and given that it's still a bit of work to do this feature properly I won't get to it very soon.

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

borkdude:
Related: https://github.com/borkdude/clj-kondo/issues/229#issuecomment-498323457

aiba:
I would find this really useful.  A common challenge for me with `clean-ns` is when I require a clojure namespace that contains a `defrecord`, and then also import that record.  As in:

```
(ns foo.core
  (:require [bar.core])
  (:import [bar.core.SomeRecord]))
```

The require is needed for the runtime generation of `SomeRecord`, but gets removed by `clean-ns`.

slipset:
[Here](https://github.com/borkdude/clj-kondo/issues/241#issuecomment-501693768) is the solution that was chosen for clj-kondo, which I think is a very wise choice. 

Any require that doesn't either use `:refer` or `:as` is assumed to be there for sideeffecting reasons and thus cannot be removed.

PR #193, opened 2017-02-24
Fix reflection warnings

julienfantin:
This PR is a very low-hanging fruit as far as performance optimizations go. It simply adds type hints to fix all the reflection warnings I could find in the `src/` tree, with only a few exceptions where a file path can be either a `String` or a `java.util.File`. Maybe the `fs` library could be used more liberally for that?

I've also added `*warn-on-reflection*` in project.clj to avoid these in the future.

Unfortunately I do not have any benchmarks to showcase performance improvements. However this branch does allow me to run commands such as `cljr-rename-symbol` on a large project whereas master consistently times out :man_shrugging:

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -7,7 +7,7 @@
             [me.raynes.fs :as fs]
             [refactor-nrepl.util :refer [normalize-to-unix-path]]
             [refactor-nrepl.s-expressions :as sexp])
-  (:import [java.io FileReader PushbackReader StringReader]))
+  (:import [java.io File FileReader PushbackReader StringReader]))
 
 (defn version []
   (let [v (-> (or (io/resource "refactor-nrepl/refactor-nrepl/project.clj")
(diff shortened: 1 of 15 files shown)


benedekfazekas:
wow, this looks really nice!

src/refactor_nrepl/core.clj
```diff
@@ -192,7 +192,7 @@
 (defn strip-reader-macros
   "Strip reader macros like #' and . (as in '(Date.)') from
   symbol-or-string."
-  [symbol-or-string]
+  ^String [symbol-or-string]
```
expez: What happens if this is a symbol?  Explosion?

expez:
Great work!  The speedup sounds significant!  

Getting an informal benchmark might not be that hard.  Could just run the `find-symbol` function from the REPL in a large project and use `time`.

src/refactor_nrepl/core.clj
julienfantin: No, this hint applies to the function's return value which is always a string

julienfantin: No, this hint applies to the function's return value which is always a string

julienfantin:
@expez that sounds easy enough but one thing I wasn't able to figure out is how to hack on the refactor nrepl backend while it's loaded in another project?

expez:
@julienfantin The quick and dirty way is to just open the jar and re-evaluate the function in question with <kbd>C-c C-c</kbd>

In this case you probably don't have to do any re-evaluation, actually.  Just change to the right namespace and run e.g. `(time (find-symbol ...))`

src/refactor_nrepl/core.clj
expez: Ah, right!

julienfantin:
Ok here are the results for a couple invocations of `find-symbol` in the project I mentioned:

Master:
```
"Elapsed time: 35208.666395 msecs"
"Elapsed time: 38600.77289 msecs"
```

This PR:
```
"Elapsed time: 13636.054625 msecs"
"Elapsed time: 12988.360079 msecs"
```

expez:
Hot damn, 3x speedup!  Fantastic! :fireworks:

expez:
New snapshot is out.  Thanks, @julienfantin!

julienfantin:
Great, thank you!

Issue #167, opened 2016-07-29
find-symbol searches all clj files in project, not just source files


gfredericks:
## Expected behavior

Only source files on the classpath should be analyzed
## Actual behavior

All `.clj` and `.cljc` files are analyzed whether or not they are source code; I also have a second checkout of my project in a `git worktree` directory and it analyzes all of those files as well.
## Steps to reproduce the problem

```
# in the root of the project
echo "#foo bar" > baz.clj
```

Try to do a replace-symbol refactoring.

A `No reader function for tag foo` error is reported.
### clj-refactor.el and refactor-nrepl version information

clj-refactor 2.2.0
### CIDER version information

```
;; CIDER 0.12.0 (Seattle), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.7.0_80
```
### Leiningen or Boot version

Leiningen 2.6.1 on Java 1.7.0_80 Java HotSpot(TM) 64-Bit Server VM
### Emacs version

24.5.1
### Operating system

Ubuntu 14.04

expez:
I don't know what you mean by a second checkout in a `git worktree`, but if the folders with the files you want excluded reside below a directory on the classpath then we're going to analyze them.

If you have some suggestions as to how we can distinguish source files we should analyze from source files we should skip, other than whether they are on the cp or not, I'm open to suggestions.

gfredericks:
No I'm not talking about files on the classpath -- the current behavior is that all files reachable from the project root are analyzed, classpath or not. The specific example I gave is a file at the root of the project.

I suppose I should distinguish between "analyzing" and "reading", because my most specific problem is that all of the files are read, which causes the `no reader function` error.

expez:
Thanks @gfredericks.  This is definitely a bug.  

@benedekfazekas The problem is the fallback dir used [here](https://github.com/clojure-emacs/refactor-nrepl/blob/master/src/refactor_nrepl/find/find_symbol.clj#L133), which is the project root.

gfredericks:
When I was doing my own debugging I *thought* I saw that that was set earlier actually, perhaps in emacs. My temporary workaround was to monkeypatch `find-symbol` so that it adds `/src` and `/test` to the `:dir` option passed in, so I'm 96% sure it's set elsewhere.

gfredericks:
[...]

PR #196, opened 2017-04-22
Add config for ignoring specific paths

naiquevin:
This change will allow users to configure refactor-nrepl to ignore
certain paths when finding the dirs on classpath by specifying a seq of
regexes which the paths will be matched against.

Currently refactor-nrepl considers all the .clj files that are reachable
from project root. This is a problem if the project repository includes
example .clj files, if a user has symlinked other clj projects under
'checkouts' or if there are any other files that need to be ignored.

This change is backward compatible and wouldn't affect the current
behaviour in existing installations as the default value is an empty
seq.

This PR addresses the issue #167. 
Will add tests after having the maintainers' thoughts on this approach. 

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### New features
 
 * New config setting `:libspec-whitelist` which makes it possible to create a seq of namespaces `clean-ns` shouldn't prune.  This is useful for libspecs which aren't used except through side-effecting loads.
+* New config setting `:ignore-paths` for ignoring certain paths when finding dirs on classpath.
 
 ## 2.2.0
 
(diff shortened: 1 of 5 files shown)


expez:
Thanks for taking an interest!

This looks like a good idea to me! :+1:

benedekfazekas:
ping @naiquevin 
planning a release soonish. do want to wrap this up so it can be part of the release? cheers..

naiquevin:
@benedekfazekas Somehow missed the notification. I will add tests in a couple of days. Thanks.

benedekfazekas:
:+1:

benedekfazekas:
build fails because mranderson prefixes deps during the build so the test ns does not find `clojure.java.classpath`. any chance you can avoid using these dependency nses in the test?

naiquevin:
Pushed a single squashed commit modifying the code and tests such that `clojure.java.classpath` is not required in the test namespace.

benedekfazekas:
ta @naiquevin

benedekfazekas:
new 2.3.0-SNAPSHOT is deployed to clojars

PR #198, opened 2017-06-24
delay loading of middleware to speed up startup

ryfow:
There is probably a more elegant way to delay code loading, but this is the most obvious way I could think of.

These changes drop somewhere between 3 and 5 seconds off of repl startup time for me. Tests pass, but I only did a test of a couple features in emacs by hand. I don't actually use refactor-nrepl very much, so I'm not particularly familiar with how people use it.

clojure-emacs/cider#1717

Thanks!

### DIFF
--- a/src/refactor_nrepl/middleware.clj
+++ b/src/refactor_nrepl/middleware.clj
@@ -4,24 +4,15 @@
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.transport :as transport]
-            [refactor-nrepl.analyzer :refer [warm-ast-cache]]
-            [refactor-nrepl.artifacts
-             :refer
-             [artifact-list artifact-versions hotload-dependency]]
             [refactor-nrepl.config :as config]
             [refactor-nrepl.core :as core]
-            [refactor-nrepl.extract-definition :refer [extract-definition]]
-            [refactor-nrepl.find.find-locals :refer [find-used-locals]]
-            [refactor-nrepl.find.find-symbol :refer [find-symbol]]
-            [refactor-nrepl.find.find-used-publics :refer [find-used-publics]]
-            [refactor-nrepl.find.find-macros :refer [warm-macro-occurrences-cache]]
-            [refactor-nrepl.ns.clean-ns :refer [clean-ns]]
             [refactor-nrepl.ns.libspecs :refer [namespace-aliases]]
-            [refactor-nrepl.ns.pprint :refer [pprint-ns]]
-            [refactor-nrepl.ns.resolve-missing :refer [resolve-missing]]
-            [refactor-nrepl.rename-file-or-dir :refer [rename-file-or-dir]]
             [refactor-nrepl.stubs-for-interface :refer [stubs-for-interface]]
             [clojure.walk :as walk]))
+
+(defn- require-and-resolve [sym]
+  (require (symbol (namespace sym)))
+  (resolve sym))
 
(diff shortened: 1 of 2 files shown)


benedekfazekas:
this deffo looks interesting. thanks for investing the energy. give me a bit more time though to have a proper look.

ryfow:
For sure. There are a variety of ways to crack this nut, and I'm quite certain that it's worth cracking. I'd love to get `cider-jack-in` to start up faster. cider-nrepl recently merged some similar commits related to cljfmt, fipp and puget being slow to load.

I've thought about writing a `delayed-refer` macro that takes arguments similar to `(:require)` and produces functions that do basically what `require-and-resolve`. It'd basically hide the fact that `(delay)` is being used under the covers. If you'd use something like that, I'd be willing to write it. Let me know if there's some other route you're interested in.

I'm not sure what other people are using for profiling load-time related stuff, but I did publish the tool I've been using. It might help you see how I got to these changes. https://github.com/ryfow/flame-du-jour

expez:
This looks good to me.

> I've thought about writing a `delayed-refer` macro that takes arguments similar to `(:require)` and produces functions that do basically what require-and-resolve does. 

Initially I was a bit put-off by having to do this for every function in the middleware ns, but this is probably the namespace we visit the least (as it's only really edited when we add new ops).

If @benedekfazekas signs off, I'm cool with merging this without spending more time on bikeshedding to make it pretty :)

benedekfazekas:
sorry for the delay (pun intended). LGTM

src/refactor_nrepl/middleware.clj
```diff
@@ -74,65 +65,112 @@
...
 (defn- find-used-locals-reply [{:keys [transport] :as msg}]
-  (reply transport msg :used-locals (find-used-locals msg)))
+  (reply transport msg :used-locals (@find-used-locals msg)))
 
 (defn- version-reply [{:keys [transport] :as msg}]
   (reply transport msg :status :done :version (core/version)))
 
+(def ^:private warm-ast-cache
+  (delay
+   (require-and-resolve 'refactor.nrepl.analyzer/warm-ast-cache)))
```
vmfhrmfoaj: I think there is typo error.

`refactor.nrepl.analzyer` => `refactor-nrepl.analzyer`

benedekfazekas:
thanks for spotting this. fortunately have not published a new snapshot yet. will check features locally before a I do.

benedekfazekas:
new snapshot is on clojars

bbatsov:
@benedekfazekas Wouldn't it be better to leverage the new API introduced by @vspinu in 0.6? I think the resulting code is cleaner and it'd be good if there's consistency between cider-nrepl and refactor-nrepl.

benedekfazekas:
yes, you are right. will try to find time to look into this. perhaps worth cutting a release too

benedekfazekas:
had a really quick look of commits/PRs around this. not sure what is the API and 0.6 you are referring to @bbatsov ... any pointer would be appreciated ;)

bbatsov:
@benedekfazekas This is the relevant commit https://github.com/clojure-emacs/cider-nrepl/commit/48dc5d24b0b20b637ba14b459549b53e5dbf5280

bbatsov:
The API in effect is two macros - https://github.com/clojure-emacs/cider-nrepl/commit/48dc5d24b0b20b637ba14b459549b53e5dbf5280#diff-4bbaa7a85d7bf72817b39b7a0452456fR17 and the one that follows.

The important thing is to just move all the middleware registration to the same file just as we did in this commit, to allow for them to be known but dynamically loaded.

PR #203, opened 2017-07-13
Fix invalid stale-check

juergenhoetzel:
The neg? check always evaluated to false because the arguments
of the time difference calculation were interchanged.

- [x ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x ] All tests are passing (run `lein do clean, test`)

### DIFF
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -38,7 +38,7 @@
 (defn- stale-cache? []
   (or (empty? @artifacts)
       (if-let [last-modified (some-> artifacts meta :last-modified .getTime)]
-        (neg? (- millis-per-day (- last-modified (.getTime (java.util.Date.)))))
+        (neg? (- millis-per-day (- (.getTime (java.util.Date.)) last-modified)))
         true)))
 
 (defn- get-mvn-artifacts!


expez:
Hi

Thanks for taking an interest, but this change doesn't look right to me.

```clj
(neg? (- millis-per-day (- (.getTime (java.util.Date.)) last-modified)))
```

If we we evaluate this right after the cache has been updated we get:

```clj
(neg? (- millis-per-day 0))
```

Which is always true.  Which means the cache is stale right after being updated.

I think what we want is:

```clj
(pos? (- millis-per-day (- (.getTime (java.util.Date.) last-modified))
```

This is false right after update, and turns true as the gap between `now` and `last-modified` grows to the size of one day.

It took me a while to figure this one out!  It would've been quite a bit easier with `clj-time`!  Unfortunately we want to try to limit the amount of dependencies.

juergenhoetzel:
Thanks for your quick feedback! :+1: 

> If we we evaluate this right after the cache has been updated we get:
> 
> (neg? (- millis-per-day 0))
> Which is always true. Which means the cache is stale right after being updated.

Actually this is `false`  right after the update, because it evaluates to `(neg? millis-per-day)` .

*After* one day `+` one millisecond the whole expression will evaluate to `(neg? -1)` which is `true`, so the cache will be stale.

> It took me a while to figure this one out! 

Exactly the same thing happened to me as to you. :tired_face:

expez:
> Actually this is false right after the update, because it evaluates to `(neg? millis-per-day)`

Try it in the REPL :)

expez:
Took my own advice and you are right.  Sorry about wasting your time on this back and forth!

Thanks for the patch! :+1:

benedekfazekas:
nice one guys! need to sort out a bug with the previous PR so don't publish to clojars yet pls

expez:
> need to sort out a bug with the previous PR so don't publish to clojars yet pls

:+1:

Issue #217, opened 2018-04-05
NullPointerException for deps.edn based projects


mbuczko:
I'm trying to prepare a ground for handling deps.edn - based projects (https://clojure.org/guides/deps_and_cli) and first thing I noticed was a NPE caused by `refactor-nrepl.core/build-artifact?` when such a project is loaded.

I prepared a simple fix which looks first if incoming `path-or-file` argument is not nil. That's pretty enough to prevent from NPE.

mbuczko:
As commented in PR, it's not a problem of NPE being thrown but rather of why `path-or-file` comes in as nil.

PR #218, opened 2018-04-05
#217 NullPointerException in refactor-nrepl.core/build-artifact? fixed

mbuczko:
fixes NPE when a deps.edn based project is loaded.

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -87,15 +87,16 @@
            :else (recur (fs/parent f))))))))
 
 (defn build-artifact? [path-or-file]
-  (let [f (io/file path-or-file)
-        target-path (-> path-or-file project-root .getCanonicalPath
-                        normalize-to-unix-path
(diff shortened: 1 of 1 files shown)


expez:
I don't mean to discourage your efforts by saying this, but getting a NPE here is the right thing.  

Calling a function that expects a `path-or-a-file` with `nil` indicates a problem elsewhere in the code.  I'd rather treat the disease than a specific symptom.

mbuczko:
Sure, thanks for commenting this out. I will try to dig a bit deeper to see where the actual problem is.

mbuczko:
@expez  you were right. Looks like it's rather the problem of `refactor-nrepl.core/project-root` which does not take _deps.edn_ into consideration as a project root. When added to the `names-at-root`, no additional checks for nil-ed `path-or-a-file` is required.

I'm closing this issue and PR and moving conversation to the new one :)

expez:
Good stuff! :+1:

PR #220, opened 2018-04-06
[Fix #219] deps.edn as recognizable project root

### LINKED ISSUE #219: deps.edn recognized as project root
As mentioned before I'd like to handle new-style dependencies (https://clojure.org/guides/deps_and_cli) introduced recently by clojure core team. The first step is to recognize _deps.edn_ as a project root (as it's already done with _project.clj_, _build.boot_ and _pom.xml_).

My PR simply adds _deps.edn_ as one more recognizable project root and this way avoids NullPointerException when deps.edn-based project is loaded.

mbuczko:
deps.edn added as a one more recognizable project root.


benedekfazekas:
LGTM

PR #222, opened 2018-05-12
Bump java.classpath

SevereOverfl0w:
0.3.0 adds support for Java 9. Apparently there are "issues" on 9.


bbatsov:
I think they should probably use the orchard.classpath instead. Also be careful about the nature of the fix - it actually does something only if you don't pass a custom classloader (which we normally need to do for boot projects).

SevereOverfl0w:
https://github.com/clojure-emacs/refactor-nrepl/blob/master/src/refactor_nrepl/core.clj#L49-L59 it does work out safe in this case, as a boot project is detected. I agree that orchard.classpath would be a better fix here.

SevereOverfl0w:
I've had confirmation that this does fix refactor-nrepl for Java 10.

SevereOverfl0w:
Superseded by #223

PR #223, opened 2018-05-13
Use orchard for classpath information

SevereOverfl0w:
Orchard has encapsulated the boot information already, so that code can
go. Orchard has already upgraded to support Java9+ too. Additionally,
orchard will receive more long-term maintenance of this code.


expez:
Thanks, this looks like a step in the right direction for sure! :+1:

bbatsov:
👍 

We should at some point think also about stuff from `refactor-nrepl` that would be useful for other other tools and can be moved to `orchard` - things like the ns cleanup and hotloading deps come to mind for instance.

expez:
Definitely @bbatsov, the more people have access to good tooling the better :)

PR #241, opened 2019-02-18
Fix a NullPointerException on jdk11

aiba:
JDK11 doesn't have system properties "sun.boot.class.path" or "java.ext.dirs".  This fix avoids a null pointer exception in `refactor-nrepl.ns.slam.hound.search`.


bbatsov:
Thanks!

PR #242, opened 2019-03-20
CircleCI migration

shen-tian:
I added a Makefile, in attempt to bring this more in line with how the other Clojure projects in the org works. Very similar CircleCI config.

https://circleci.com/gh/shen-tian/workflows/refactor-nrepl

As it stood, the project was tested on Oracle Java 8 and Clojure 1.8 and 1.9 in a single job. I've move that over to OpenJDK 8, added Clojure 1.10 tests, and split them up into separate jobs.

I've tried to add Java 11 tests, but they are not passing. Will jot down some notes so we can investigate. Given that we are not testing on Java 11 right now, this PR is good to go.

Edit: a quick look suggests that #241 addresses the Java 11 issues, as that function is where the tests are failing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,11 +116,14 @@ jobs:
           cache_version: << parameters.clojure_version >>
           steps:
             - run:
-                name: Installing source-deps
-                command: lein source-deps :prefix-exclusions "[\"classlojure\"]"
-            - run:
                 name: Running tests
-                command: lein with-profile +<< parameters.clojure_version >>,+plugin.mranderson/config test
+                command: make test
+            # - run:
+            #     name: Installing source-deps
+            #     command: lein source-deps :prefix-exclusions "[\"classlojure\"]"
+            # - run:
+            #     name: Running tests
+            #     command: lein with-profile +<< parameters.clojure_version >>,+plugin.mranderson/config test
 
 ######################################################################
 #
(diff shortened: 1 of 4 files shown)


benedekfazekas:
thanks for this. will have a proper look shortly

Makefile
```diff
@@ -0,0 +1,29 @@
+.PHONY: source-deps test release deploy clean
+
+VERSION ?= 1.9
```
bbatsov: Shouldn't this be 1.10?

shen-tian: That's a copy+paste. I assume that's the default value? If so, will change to 1.10

.circleci/config.yml
```diff
@@ -0,0 +1,192 @@
...
+      - with_cache:
+          cache_version: << parameters.clojure_version >>
+          steps:
+            - run:
+                name: Downloading Source deps
+                command: make source-deps
+            - run:
+                name: Running tests
+                command: make test
+            # - run:
```
bbatsov: You can remove this, as I see it's in the Makefile already.

bbatsov:
Apart from my small remarks everything looks great to me.

Makefile
shen-tian: Fixed

.circleci/config.yml
shen-tian: Removed

shen-tian:
Ok. I just addressed the comments, rebased to include #241, and switched Java 11 tests on.

bbatsov:
Fantastic work! Thank you! 🙇

PR #243, opened 2019-03-20
Adding cljfmt check

shen-tian:
Uses the same config as `cider-nrepl`. Also runs cljfmt to fix things that came up.

Splitting the commits to 1) adding the profile/CI job, and 2) actually fixing the formatting.


bbatsov:
Thanks!

PR #244, opened 2019-03-25
Upgrade mranderson to 0.5.0

benedekfazekas:
mranderson 0.5.x specific config added to the project file as
well. Using resolved tree mode. This means that
mranderson works with a resolved dependency tree and does not create a
deeply nested directory structure, only adds prefixes to all
dependencies. Handles all depedendencies as first level ones and
processes them in a topological order.

Add project prefix in the project file and other config for unresolved
tree mode in case we want to switch to that in the near future

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -7,14 +7,10 @@
                  ^:source-dep [http-kit "2.3.0"]
                  ^:source-dep [cheshire "5.8.0"]
                  ^:source-dep [org.clojure/tools.analyzer.jvm "0.7.1"]
-                 ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha3"]
-                 ;; Not used directly in refactor-nrepl, but needed because of tool.namespace
-                 ;; and the way MrAnderson processes dependencies
-                 ;; See https://github.com/clojure-emacs/cider/issues/2176 for details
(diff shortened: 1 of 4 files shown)


project.clj
```diff
@@ -38,7 +37,7 @@
              :1.10 {:dependencies [[org.clojure/clojure "1.10.0"]
                                    [org.clojure/clojurescript "1.10.63"]]}
              :dev {:plugins [[jonase/eastwood "0.2.0"]]
-                   :global-vars {*warn-on-reflection* true}
+                   ;:global-vars {*warn-on-reflection* true}
```
expez: oops?

project.clj
```diff
@@ -23,7 +19,10 @@
                                     :username :env/clojars_username
                                     :password :env/clojars_password
                                     :sign-releases false}]]
-  :plugins [[thomasa/mranderson "0.4.8"]]
+  :plugins [[thomasa/mranderson "0.5.0"]]
+  :mranderson {:project-prefix  "refactor.inlined-deps"
```
expez: Perhaps `refactor-nrepl.inlined-deps` would be better?

benedekfazekas:
cheers, fixes pushed

PR #247, opened 2019-03-28
Parallelise `refactor-nrepl.find.find-symbol/find-global-symbol`

vemv:
(Second attempt after the failed https://github.com/clojure-emacs/refactor-nrepl/pull/246 )

## What

`refactor-nrepl.find.find-symbol/find-symbol-in-file` is an expensive op, particularly because it can trigger an AST analysis for a given ns.

That, combined with the large workload of analysing [all](https://github.com/clojure-emacs/refactor-nrepl/blob/bb27d1345cc564aa8f3b5ecad7c0560f221a3e0b/src/refactor_nrepl/find/find_symbol.clj#L138) files within a given project seems to justify parallelization.

## Benchmarks

Applying [this patch](https://gist.github.com/vemv/300582528b1839e450d68160c1e865cf) to my branch will add a poor man's benchmark. With it, each clj-refactor.el `rs` will use map or pmap intermittently, and log the time:

Screenshot 2019-03-28 at 14 47 13

(the screenshot was taken after multiple 'warmup' runs)

---

  * build.sh fails for me even on latest `master`
  * But the changes there are quite unlikely to be build-breaking.
  * (Impl detail)
  * (Impl detail)

### DIFF
--- a/src/refactor_nrepl/find/find_symbol.clj
+++ b/src/refactor_nrepl/find/find_symbol.clj
@@ -140,7 +140,7 @@
          (find-util/distribute-evenly-by {:n (find-util/processor-count)
                                           :f (fn [^File file]
                                                (-> file .length))})
-         (find-util/divide-by (find-util/processor-count))
+         (find-util/slice (find-util/processor-count))
          (pmap (fn [work]
                  (->> work
                       (mapcat (comp doall
--- a/src/refactor_nrepl/find/find_symbol.clj
+++ b/src/refactor_nrepl/find/find_symbol.clj
@@ -140,13 +140,12 @@
          (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
          (find-util/divide-by (find-util/processor-count))
          (pmap (fn [work]
-                 ;; `vec` ensures all work is actually performed within the surrounding `pmap`
                  (->> work
-                      (mapcat (comp vec
+                      (mapcat (comp doall
                                     (partial find-symbol-in-file fully-qualified-name ignore-errors)))
-                      vec)))
+                      doall)))
          (apply concat)
-         (vec))))
+         (doall))))
 
 (defn- get&read-enclosing-sexps
   [file-content {:keys [^long line-beg ^long col-beg]}]
--- a/src/refactor_nrepl/find/find_symbol.clj
+++ b/src/refactor_nrepl/find/find_symbol.clj
@@ -137,7 +137,16 @@
                                (str/join "/" [namespace var-name]))]
     (->> (core/dirs-on-classpath)
          (mapcat (partial core/find-in-dir (some-fn core/clj-file? core/cljc-file?)))
-         (mapcat (partial find-symbol-in-file fully-qualified-name ignore-errors)))))
+         (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
+         (find-util/divide-by (find-util/processor-count))
+         (pmap (fn [work]
+                 ;; `vec` ensures all work is actually performed within the surrounding `pmap`
+                 (->> work
+                      (mapcat (comp vec
+                                    (partial find-symbol-in-file fully-qualified-name ignore-errors)))
+                      vec)))
+         (apply concat)
+         (vec))))
 
 (defn- get&read-enclosing-sexps
   [file-content {:keys [^long line-beg ^long col-beg]}]
--- a/src/refactor_nrepl/find/find_symbol.clj
+++ b/src/refactor_nrepl/find/find_symbol.clj
@@ -137,7 +137,9 @@
                                (str/join "/" [namespace var-name]))]
     (->> (core/dirs-on-classpath)
          (mapcat (partial core/find-in-dir (some-fn core/clj-file? core/cljc-file?)))
-         (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
+         (find-util/distribute-evenly-by {:n (find-util/processor-count)
+                                          :f (fn [^File file]
+                                               (-> file .length))})
          (find-util/divide-by (find-util/processor-count))
          (pmap (fn [work]
                  (->> work
--- a/src/refactor_nrepl/find/util.clj
+++ b/src/refactor_nrepl/find/util.clj
@@ -27,5 +27,28 @@
   (let [c (/ (count coll) n)]
     (partition-all c coll)))
 
+(defn distribute-evenly-by
+  "Sorts `coll` by `f` in such a way that if partitioned by `n`, each partition will have items of similar cost."
+  [{:keys [f n]
+    :as   options}
+   coll]
+  {:pre [(partial every? #{:f :n}) (keys options)
+         (ifn? f)
+         (pos? n)
+         (integer? n)
+         (coll? coll)]}
+  (if-not (seq coll)
+    coll
+    (->> coll
+         (sort-by f)
+         (partition-all n)
+         (map (fn [chunk]
+                (->> (repeat ::padding)
+                     (concat chunk)
+                     (take n))))
+         (apply map vector)
+         (apply concat)
+         (remove #{::padding}))))
+
 (defn processor-count []
   (-> (Runtime/getRuntime) .availableProcessors))
--- a/src/refactor_nrepl/find/util.clj
+++ b/src/refactor_nrepl/find/util.clj
@@ -20,3 +20,12 @@
              (core/suffix name)))))
   ([file occ]
    (spurious? (assoc occ :file file))))
+
+(defn divide-by
+  "Divides `coll` in `n` parts. The parts can have disparate sizes if the division isn't exact."
+  [n coll]
+  (let [c (/ (count coll) n)]
+    (partition-all c coll)))
+
+(defn processor-count []
+  (-> (Runtime/getRuntime) .availableProcessors))
--- a/src/refactor_nrepl/find/util.clj
+++ b/src/refactor_nrepl/find/util.clj
@@ -21,7 +21,7 @@
   ([file occ]
    (spurious? (assoc occ :file file))))
 
-(defn divide-by
+(defn slice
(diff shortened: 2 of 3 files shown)


src/refactor_nrepl/find/find_symbol.clj
```diff
@@ -137,7 +137,16 @@
...
+         (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
+         (find-util/divide-by (find-util/processor-count))
+         (pmap (fn [work]
+                 ;; `vec` ensures all work is actually performed within the surrounding `pmap`
+                 (->> work
+                      (mapcat (comp vec
+                                    (partial find-symbol-in-file fully-qualified-name ignore-errors)))
+                      vec)))
+         (apply concat)
+         (vec))))
```
vemv: One might remove this `vec`: art of the point of `pmap` is lazyness.

In `defn find-symbol` at the bottom of this file, `@globals` is the last option in the `or`, so it might make sense to keep it lazy.

OTOH keeping the `(vec)` might be desirable so one can benchmark things without having to remember about seq realisation.

src/refactor_nrepl/find/util.clj
```diff
@@ -20,3 +20,12 @@
              (core/suffix name)))))
   ([file occ]
    (spurious? (assoc occ :file file))))
+
+(defn divide-by
```
expez: How about `chunk`?  With a docstring of "Split `coll` into `n` chunks.  The final chunk might be smaller than the rest."

src/refactor_nrepl/find/find_symbol.clj
```diff
@@ -137,7 +137,16 @@
...
          (mapcat (partial core/find-in-dir (some-fn core/clj-file? core/cljc-file?)))
-         (mapcat (partial find-symbol-in-file fully-qualified-name ignore-errors)))))
+         (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
+         (find-util/divide-by (find-util/processor-count))
+         (pmap (fn [work]
+                 ;; `vec` ensures all work is actually performed within the surrounding `pmap`
+                 (->> work
+                      (mapcat (comp vec
+                                    (partial find-symbol-in-file fully-qualified-name ignore-errors)))
+                      vec)))
```
expez: Btw I think `doall` signals much better that the intention is to realize a lazy value.

use of `vec` is likely to get refactored away when future selves or other developers don't see the point of conversion to this data type.

This can be changed in a few places in this diff.

src/refactor_nrepl/find/find_symbol.clj
```diff
@@ -137,7 +137,16 @@
                                (str/join "/" [namespace var-name]))]
     (->> (core/dirs-on-classpath)
          (mapcat (partial core/find-in-dir (some-fn core/clj-file? core/cljc-file?)))
-         (mapcat (partial find-symbol-in-file fully-qualified-name ignore-errors)))))
+         (shuffle) ;; make it less likely that work will be concentrated in one partition (see next line)
```
expez: Can't see how that's true.  

The random shuffled seq has no better guarantees of an even split than whatever order is returned when we read from disk.  It's just be different each time.

vemv: Assume 2 cores and the following workload:

`[:tiny, :tiny, :heavy, :heavy]`

...where `:heavy` takes 2x to process.

Without shuffling, cores would get an uneven distribution 100% of the times, and the overall task performance will be bound by the busiest core.

I would say this is a valid analogy for real-world projects. Those tend to be organised in directories with different purposes and authors, so there's a strong chance that their files will have disparate sizes (and therefore costs).

src/refactor_nrepl/find/find_symbol.clj
vemv: > Let's leave it as eager or the value won't be calculated in the background.

Nicely summarised. Didn't realise!

src/refactor_nrepl/find/util.clj
vemv: Can do!

However:

imagine you see `(chunk coll 3)` in the wild, without further context. Would you imagine/remember it _divides_ the members by 3, rather than returning chunks _of size_ 3 (which is different)?

src/refactor_nrepl/find/find_symbol.clj
vemv: That said one _could_ use a heuristic for attempting a deterministic, even distribution, by querying each file's `.length()` in bytes.

vemv:
Force-pushed, I had left an outdated comment hanging.

src/refactor_nrepl/find/util.clj
expez: > Would you imagine/remember it divides the members by 3, rather than returning chunks of size 3 (which is different)?

This part really confuses me.  I thought maybe I had misunderstood what `divide-by` did, but one of the tests:

```clj
(chunk [1 1 1] 3)
;;=> '((1) (1) (1))
```
Is exactly how I'd expect chunking to work: it creates 3 equally sized chunks of the input seq.

I agree that the name isn't perfect (could work exactly like `partition`), I just thought `divide-by` gave little indication of what went on.  Perhaps we can come up with something else entirely?  `partition-evenly`?  I mean it's just like partition except in the degenerate case where you get one-element partitions instead of a single partition with everything.

src/refactor_nrepl/find/find_symbol.clj
expez: > I would say this is a valid analogy for real-world projects. Those tend to be organised in directories with different purposes and authors, so there's a strong chance that their files will have disparate sizes (and therefore costs).

Don't agree with this assumption.  Most organizations have a common code style, e.g. either all files are allowed to grow long, or none are.  If all files are allowed to grow long they'll probably be scattered randomly throughout the project as a byproduct of task complexity.

My objection was to doing extra work and introducing non-determinism (users are left wondering why the op is sometimes fast and other times slow) with no guarantee of a better outcome.

For all we know we read `[:tiny, :heavy, :tiny, :heavy]` from disk and then shuffle it into `[:tiny, :tiny, :heavy, :heavy]`.

Thanks for explaining your thinking.  Since the op was down from 1750ms to 350ms going in either direction is fine with me.

vemv: Valid objections :) thanks as well!

The last suggestion (`.length`) might be spot on though.

src/refactor_nrepl/find/util.clj
vemv: > Is exactly how I'd expect chunking to work: it creates 3 equally sized chunks of the input seq.

Not sure I agree here. clojure.core chunks have a different meaning, which influences my thought.

Perhaps `in-groups-of` would do? Name picked after https://apidock.com/rails/Array/in_groups_of which does the same thing

src/refactor_nrepl/find/util.clj
```diff
@@ -20,3 +20,12 @@
...
+  "Divides `coll` in `n` parts. The parts can have disparate sizes if the division isn't exact."
+  [n coll]
+  (let [c (/ (count coll) n)]
+    (partition-all c coll)))
+
+(defn processor-count []
+  (-> (Runtime/getRuntime) .availableProcessors))
```
thumbnail: you might want to configure this [for the same reason eftest has it as a configuration option](https://github.com/weavejester/eftest/blob/master/README.md#setting-the-number-of-threads-used)

src/refactor_nrepl/find/util.clj
expez: I like that! 👍

vemv: Argh, I lied:

```
pry> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].in_groups_of(3)
=> [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]]
```

`in_groups_of` is exactly like `partition`, with no 'division' semantics. Not good to use the identical vocabulary for different things (which is my objection to `chunk` as well)

vemv:
Let me know how should we resolve the two outstading items. 100% willing to follow the original suggestions, but maybe I made some valid points ;p

src/refactor_nrepl/find/util.clj
expez: I tried to look for prior art:  This function exists in flatland as [slice](https://cljdoc.org/d/org.flatland/useful/0.11.6/api/flatland.useful.seq#slice).

I'm not excited by that name, but I think it's better than `divide-by`.  FWIW I would've inlined this function a long time ago after failing to find a good name :p

src/refactor_nrepl/find/find_symbol.clj
expez: To me it makes sense to either remove the `shuffle` or to use the `.length` as a heuristic to re-order the collection.

I agree that it's worth re-ordering the sequence, in the sense that the time it takes to do so will be more than repaid by better parallelization.

vemv:
Feedback processed, let me know what you think!

expez:
This looks good now, but what's with the test failure?

vemv:
Had noticed, seemed spurious. I also get sporadic failures locally.

expez:
@bbatsov Have you seen this before after the migration to circleci?  I tried re-running the build but got the same outcome.

vemv:
Any luck with this one?

Perhaps a CI retrigger would work by now

expez:
Tried rebuilding and it's still failing...

vemv:
(Rebased against master)

I have the feeling that `find-symbol-in-file` is simply thread-unsafe, hence the seemingly-random failing tests.

That's pretty understandable, since evaluating Clojure code in parallel is risky (`clojure.core/require` isn't thread-safe to begin with).

Accomplishing this PR might need a pretty disciplined approach to performing the AST evaluations in a specific order ("dependency order").

That's not impossible to achieve, but meanwhile I'll close this PR as I don't see myself making that effort anytime soon.

PR #249, opened 2019-04-02
Updating the deps as far as we can without breaking tests.

shen-tian:
Hoping this is useful. I've updated the deps as far as possible without breaking tests. Of what remain:

- `cider-nrepl` 0.21 requires `nrepl` 0.6. Moving to these two broke some tests. This might be something to do with the new printing middleware.
- `cljs-tooling` 0.3 changed the varity of a function, as part of its self-hosted cljs change. This looks fixable.

---

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ install.cmd
 /.cljs_node_repl/
 /out/
 /.lein-env
+.inline-deps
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

> cider-nrepl 0.21 requires nrepl 0.6. Moving to these two broke some tests. This might be something to do with the new printing middleware.

What exactly got broken? I guess it won't be hard to fix.

> cljs-tooling 0.3 changed the varity of a function, as part of its self-hosted cljs change. This looks fixable.

Everything is fixable. ;-)

PR #251, opened 2019-04-18
clean-ns: fall back to :relative-path, when given

plexus:
Allow the client to send both the absolute path and a relative path to a file,
we use the first one that we can find.

The absolute path is used when the client (Emacs) and the JVM are on the same
filesystem/same machine. The second is used when they are not on the same
filesystem (e.g. running nREPL on a remote machine or in a VM), but the JVM runs
in the (mounted) project directory, so that relative file lookups still work.

I failed to run the tests locally. Will see what CI says.

See https://github.com/clojure-emacs/clj-refactor.el/issues/380#issuecomment-484509768

### DIFF
--- a/src/refactor_nrepl/ns/clean_ns.clj
+++ b/src/refactor_nrepl/ns/clean_ns.clj
@@ -20,7 +20,8 @@
              [ns-parser :as ns-parser]
              [prune-dependencies :refer [prune-dependencies]]
              [rebuild :refer [rebuild-ns-form]]]
-            [clojure.string :as str]))
+            [clojure.string :as str]
+            [clojure.java.io :as io]))
 
 (defn- assert-no-exclude-clause
   [use-form]
@@ -35,20 +36,22 @@
   (assert-no-exclude-clause (core/get-ns-component ns-form :use))
   ns-form)
 
-(defn clean-ns [{:keys [path]}]
+(defn clean-ns [{:keys [path relative-path]}]
   {:pre [(seq path) (string? path) (core/source-file? path)]}
-  ;; Prefix notation not supported in cljs.
-  ;; We also turn it off for cljc for reasons of symmetry
-  (config/with-config {:prefix-rewriting (if (or (core/cljs-file? path)
-                                                 (core/cljc-file? path))
-                                           false
-                                           (:prefix-rewriting config/*config*))}
-    (let [ns-form (validate (core/read-ns-form-with-meta path))
-          deps-preprocessor (if (get config/*config* :prune-ns-form)
-                              #(prune-dependencies % path)
-                              identity)
-          new-ns-form (-> (ns-parser/parse-ns path)
-                          deps-preprocessor
-                          (rebuild-ns-form ns-form))]
-      (when-not (= ns-form new-ns-form)
-        new-ns-form))))
+  ;; Try first the absolute, then the relative path
+  (let [path (first (filter #(some-> % io/file .exists) [path relative-path]))]
+    ;; Prefix notation not supported in cljs.
+    ;; We also turn it off for cljc for reasons of symmetry
+    (config/with-config {:prefix-rewriting (if (or (core/cljs-file? path)
+                                                   (core/cljc-file? path))
+                                             false
+                                             (:prefix-rewriting config/*config*))}
+      (let [ns-form (validate (core/read-ns-form-with-meta path))
+            deps-preprocessor (if (get config/*config* :prune-ns-form)
+                                #(prune-dependencies % path)
+                                identity)
+            new-ns-form (-> (ns-parser/parse-ns path)
+                            deps-preprocessor
+                            (rebuild-ns-form ns-form))]
+        (when-not (= ns-form new-ns-form)
+          new-ns-form)))))
--- a/src/refactor_nrepl/ns/clean_ns.clj
+++ b/src/refactor_nrepl/ns/clean_ns.clj
@@ -37,9 +37,10 @@
   ns-form)
 
 (defn clean-ns [{:keys [path relative-path]}]
-  {:pre [(seq path) (string? path) (core/source-file? path)]}
+  {:pre [(seq path) (string? path)]}
   ;; Try first the absolute, then the relative path
   (let [path (first (filter #(some-> % io/file .exists) [path relative-path]))]
+    (assert (core/source-file? path))
     ;; Prefix notation not supported in cljs.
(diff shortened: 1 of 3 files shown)


plexus:
```
FAIL in (sanity) (form-init5821700100748902327.clj:67)
errors handled properly
expected: (string? (:err response))
  actual: (not (string? nil))
```

This seems unrelated, is this a flaky test?

src/refactor_nrepl/ns/clean_ns.clj
```diff
@@ -35,20 +36,23 @@
...
-                              identity)
-          new-ns-form (-> (ns-parser/parse-ns path)
-                          deps-preprocessor
-                          (rebuild-ns-form ns-form))]
-      (when-not (= ns-form new-ns-form)
-        new-ns-form))))
+(defn clean-ns [{:keys [path relative-path]}]
+  {:pre [(seq path) (string? path)]}
+  ;; Try first the absolute, then the relative path
+  (let [path (first (filter #(some-> % io/file .exists) [path relative-path]))]
```
expez: `some` might be better here since we're looking for the first thing that fulfills some `pred`

plexus: Since the predicate returns boolean rather than the truthy value I find it awkward to use `some` for such cases, as you end up with `(some #(and (pred? %) %) coll)`. Would you still prefer `some`?

plexus:
I've updated the CHANGELOG and README, and rewrote that use of `first` + `filter`.

It's very annoying that I can't run tests locally, no matter what I do I get

``` clojure
clojure.lang.LispReader$ReaderException: java.lang.RuntimeException: No reader function for tag Inf
```

when [mranderson loads](https://github.com/benedekfazekas/mranderson/blob/master/src/mranderson/move.clj#L26) the ns [parallel.core](https://github.com/reborg/parallel/blob/master/src/parallel/core.clj#L323). I've tried forcing the Clojure version to 1.10 but nothing seems to work.

plexus:
I'm very sorry, I accidentally pushed this to master, and I can't rewind because master is protected. I hope this looks good! I'll try to do some better testing to make sure, let me know if you think this needs more work @expez and I'll do it on a new PR.

expez:
I tried grepping the project for `##Inf` but I can't find anything extending the reader to understand that tag.  There's also no messing around with `*default-data-reader-fn*` AFAICT.

@benedekfazekas Does it make a huge difference if you use that library or can we drop it from mranderson?  If not, we might consider opening an issue over at that repo to let him know that this is causing breakage over here.

plexus:
Clojure itself understands `##Inf`, not 100% sure when it was added, I think maybe 1.9.

It also seems to work on CI so clearly I'm doing something wrong, I just have no clue what. I'm just trying to run `make test`. I've tried blowing away my `~/.m2` but that didn't help.

benedekfazekas:
yeah aware of this new limitation of MrAnderson. however, this only means that the MrAnderson step needs to be run with `lein` 2.9.1 (older leiningen version may pull an older version of clojure maybe). don't think this should cause any problems downstream, or?

benedekfazekas:
ah checked the above comments properly, guessing @plexus that you use an older version of leiningen?

plexus:
It would appear so, I was still on 2.8.3 apparently.

plexus:
Did a `lein upgrade` and now it seems to work. Can't believe I didn't check that to begin with :see_no_evil: . Thanks @benedekfazekas !

benedekfazekas:
nw @plexus, happy it is solved

jonpither:
When I do a clean-ns using the latest snapshot `2.5.0-SNAPSHOT`, on a project with check-out deps, I get:

```
cljr--get-error-value: Error in nrepl-refactor: java.lang.NullPointerException: null
 at refactor_nrepl.core$cljc_file_QMARK_.invokeStatic (core.clj:129)
    refactor_nrepl.core$cljc_file_QMARK_.invoke (core.clj:127)
    clojure.core$some_fn$sp3__8653.invoke (core.clj:7461)
    refactor_nrepl.core$source_file_QMARK_.invokeStatic (core.clj:152)
    refactor_nrepl.core$source_file_QMARK_.invoke (core.clj:146)
    refactor_nrepl.ns.clean_ns$clean_ns.invokeStatic (clean_ns.clj:44)
    refactor_nrepl.ns.clean_ns$clean_ns.invoke (clean_ns.clj:39)
    clojure.lang.Var.invoke (Var.java:384)
    refactor_nrepl.middleware$clean_ns_reply.invokeStatic (middleware.clj:134)
    refactor_nrepl.middleware$clean_ns_reply.invoke (middleware.clj:133)
    refactor_nrepl.middleware$wrap_refactor$fn__6066.invoke (middleware.clj:212)
    nrepl.middleware$wrap_conj_descriptor$fn__630.invoke (middleware.clj:16)
```

jonpither:
Updated to the latest clj-refactor.el, now I get:

```
=> nil
cljr--get-error-value: Error in nrepl-refactor: java.lang.AssertionError: Assert failed: (instance? java.io.PushbackReader rdr)
 at clojure.tools.namespace.parse$read_ns_decl.invokeStatic (parse.clj:33)
    clojure.tools.namespace.parse$read_ns_decl.invoke (parse.clj:33)
    refactor_nrepl.core$read_ns_form.invokeStatic (core.clj:108)
    refactor_nrepl.core$read_ns_form.invoke (core.clj:105)
    refactor_nrepl.core$clj_file_QMARK_.invokeStatic (core.clj:144)
    refactor_nrepl.core$clj_file_QMARK_.invoke (core.clj:139)
    clojure.core$some_fn$sp3__8653.invoke (core.clj:7461)
    refactor_nrepl.core$source_file_QMARK_.invokeStatic (core.clj:152)
    refactor_nrepl.core$source_file_QMARK_.invoke (core.clj:146)
    refactor_nrepl.ns.clean_ns$clean_ns.invokeStatic (clean_ns.clj:44)
    refactor_nrepl.ns.clean_ns$clean_ns.invoke (clean_ns.clj:39)
    clojure.lang.Var.invoke (Var.java:384)
    refactor_nrepl.middleware$clean_ns_reply.invokeStatic (middleware.clj:134)
    refactor_nrepl.middleware$clean_ns_reply.invoke (middleware.clj:133)
    refactor_nrepl.middleware$wrap_refactor$fn__6066.invoke (middleware.clj:212)
```

benedekfazekas:
that is a know issue with latest snapshot (not really related to this issue) i still have not got around to check. sorry.

PR #252, opened 2019-04-22
Use seek instead of find, the latter already exists in clojure.core

plexus:
Follow up to #251, I had completely forgotten `clojure.core` already has a `find` function.

### DIFF
--- a/src/refactor_nrepl/ns/clean_ns.clj
+++ b/src/refactor_nrepl/ns/clean_ns.clj
@@ -36,13 +36,13 @@
   (assert-no-exclude-clause (core/get-ns-component ns-form :use))
   ns-form)
 
-(defn- find
+(defn- seek
   "Find the first element in coll that satisfies pred?"
   [pred? coll]
   (reduce (fn [_ x] (when (pred? x) (reduced x))) nil coll))
 
 (defn clean-ns [{:keys [path relative-path]}]
-  (let [path (find #(and % (.exists (io/file %)))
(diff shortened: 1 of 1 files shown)


src/refactor_nrepl/ns/clean_ns.clj
```diff
@@ -36,13 +36,13 @@
...
 
-(defn- find
+(defn- seek
   "Find the first element in coll that satisfies pred?"
   [pred? coll]
   (reduce (fn [_ x] (when (pred? x) (reduced x))) nil coll))
 
 (defn clean-ns [{:keys [path relative-path]}]
-  (let [path (find #(and % (.exists (io/file %)))
+  (let [path (seek #(and % (.exists (io/file %)))
```
expez: Just use `fn` instead of the `#()` notation with `some`.  

I agree with you that it's awkward to continue expanding the `#()` but implementing `seek` as a slight variation of `some` isn't any better.  I just shifts the question from "What does this hard-to-read anonymous function do" to "what does `seek` do"?

plexus: Sure, whatever you prefer. This was prompted by some discussion on twitter, the reduce/reduced version has some benefits in terms of performance, but I agree it's overkill in this case.

https://mobile.twitter.com/plexus/status/1119957160685985793

plexus:
@expez ok like this?

expez:
Looks good to me @plexus :)

PR #253, opened 2019-04-23
Mention the minimum version requirement of Leiningen

plexus:
Hopefully it saves someone some frustration down the line (down the lein?).


benedekfazekas:
indeed, thx for this too!!

PR #256, opened 2019-06-04
Add try/catch when reading EDN data

### LINKED ISSUE #255: `cljr-add-project-dependency` blows up when parsing Clojar artifacts
Cross-posting here.

Sorry for skipping the report template - all details should be in the linked Github issue:

https://github.com/syl20bnr/spacemacs/issues/12385

Should also add that I'm running OpenJDK 8 (AdoptOpenJDK).

andersenleo:
Fixes #255 

Since the `artifacts/get-clojars-artifacts!` function is written in a way that isn't easily testable I opted out on tests. I would need to refactor the code a bit more to make it possible to write a test for this. Let me know if you'd like me to do that.

The fix is simply to use `keep` instead of `map` - and swallow any exception on `edn/read-string`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * [#251](https://github.com/clojure-emacs/refactor-nrepl/pull/251) `clean-ns` support extra message key `relative-path`, which will be used if `path` does not exist.
+* [#256](https://github.com/clojure-emacs/refactor-nrepl/pull/256) ignore malformed artifact coordinates when fetching from Clojars.
 
 ## 2.4.0
 
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -39,7 +39,11 @@
          java.net.URL.
          io/reader
          line-seq
-         (map edn/read-string))
+         (keep #(try
+                  (edn/read-string %)
+                  (catch Exception _
+                    ;; Ignore artifact if not readable. See #255
+                    nil))))
     (catch Exception _
       ;; In the event clojars is down just return an empty vector. See #136.
       [])))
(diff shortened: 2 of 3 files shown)


benedekfazekas:
A test would be nice I think. Please also add a one liner to the changelog. Thanks again, really appreciated.

andersenleo:
Moved the "reading" to a separate function and wrote a minimal test for verifying that. Let me know if you'd like to restructure things differently.

benedekfazekas:
good work @andersenleo! will release a new snapshot in the evening.

expez:
What was the point of this?  There already was a `try/catch` in place making sure the client got `[]` when clojars was down (or edn parsing failed).  What am I missing?

andersenleo:
@expez - the `edn/read-string` function is called when the lazy-seq returned from `get-clojars-artifacts!` is evaluated further down the line. The `try/catch` that was in place would only return `[]` if the connection to Clojars was down (an error triggered before call to `line-seq`). And either way -- returning an empty seq would still make the feature unusable (if the artifact seq would have been eagerly evaluated).

expez:
Thanks @andersenleo I eventually got that and deleted my comment hoping nobody would've seen it! :p 

Great job on this! 👍

benedekfazekas:
new snapshot version is on clojars

andersenleo:
Lovely - it seems to work. I'm no Emacs expert, but adding
```elisp
  (setq cider-jack-in-lein-plugins '(("refactor-nrepl" "2.4.1-SNAPSHOT" :predicate cljr--inject-middleware-p)
                                     ("cider/cider-nrepl" "0.22.0-beta1")))

```
works fine. Should be an easier way to override the refactor-nrepl version I guess? Suggestions are welcome :). Any idea when a "proper" release could be made available?

benedekfazekas:
looks legit to me. no release is planned atm. nothing blocking one either afaik

PR #261, opened 2019-06-20
Add support for TRAMP connection

notake:
- Added processing to remove the TRAMP method (e.g. `/ssh:user@host:` ) from the `file` parameter before calling `slurp`.
- Fixes https://github.com/clojure-emacs/clj-refactor.el/issues/350 .
- Fixes https://github.com/clojure-emacs/clj-refactor.el/issues/356 .

---

### DIFF
--- a/src/refactor_nrepl/extract_definition.clj
+++ b/src/refactor_nrepl/extract_definition.clj
@@ -2,6 +2,7 @@
   (:require [clojure.string :as str]
             [refactor-nrepl
              [core :refer [suffix]]
+             [tramp :as tramp]
              [s-expressions :as sexp]]
             [refactor-nrepl.find.find-symbol :refer [find-symbol]]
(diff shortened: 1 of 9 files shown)


bbatsov:
I'll take a closer look at the changes, but my initial observation is that something like this should probably be in the util ns or something like this. I also don't recall such a problem manifesting in CIDER, so probably there's a different (less invasive) way to handle the TRAMP urls.

expez:
Wouldn't it be better if clj-refactor dealt with the tramp stuff?  

My thinking is that Tramp is an Emacs thing, and refactor-nrepl is meant to be a generic backend.

notake:
I decided to create TRAMP connection support for clj-refactor.el instead of refactor-nrepl.  https://github.com/clojure-emacs/clj-refactor.el/pull/445
Thank you very much for the review.

PR #262, opened 2019-07-09
Update dependencies

nwjsmith:
I'm working on getting the tests to pass for bumps to `nrepl` and `cider-nrepl`, but in the meantime the other dependencies can be updated.

Thanks!

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -7,7 +7,7 @@
                  ^:inline-dep [http-kit "2.3.0"]
                  ^:inline-dep [cheshire "5.8.1"]
                  ^:inline-dep [org.clojure/tools.analyzer.jvm "0.7.2"]
-                 ^:inline-dep [org.clojure/tools.namespace "0.3.0-alpha3" :exclusions [org.clojure/tools.reader]]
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -29,11 +29,9 @@
              :test {:dependencies [[print-foo "1.0.2"]]
                     :src-paths ["test/resources"]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]
-                                  [org.clojure/clojurescript "1.8.51"]
-                                  [javax.xml.bind/jaxb-api "2.3.1"]]}
+                                  [org.clojure/clojurescript "1.8.51"]]}
```
bbatsov: This will causes breakages for ClojureScript. We don't use this directly, but I'm afraid it's needed.

nwjsmith:
@bbatsov oops, I opened this and walked away before seeing the test failures. I've dropped the `java.xml.bind` change, and also the update to `leiningen-core` 2.9.1 because it relies on features only available in Clojure 1.9+. All tests are passing now

Issue #263, opened 2019-07-31
Make Artifacts Cache Durable


paulrd:
Hi this is a feature request. Would it be possible to make the artifacts cache durable (write to local file) so that https://clojars.org/repo/all-jars.clj is not downloaded every time cider jacks-in? I have the bad habit of restarting the cider repl many times and that's 4MB each time and I have to pay a lot for bandwidth. Also, bandwidth could be saved by downloading the gzipped version which is a fraction (1/8th) the size (500KB) https://clojars.org/repo/all-jars.clj.gz.

paulrd:
I can give this a shot. It just needs to write a timestamp to a file and switch to using gzip.

expez:
Hi, this sounds like a good idea to me.

We can do either or both.  Do keep in mind that you have to check for the presence of gzip before running the new code path.  Especially our users on Windows are unlikely to have that utility available on path for emacs to use.

PR #264, opened 2019-08-01
Lower bandwidth at startup

paulrd:

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,3 @@ install.cmd
 /out/
 /.lein-env
 .inline-deps
-/.artifacts-cache
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ install.cmd
 /out/
 /.lein-env
 .inline-deps
+/.artifacts-cache
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -24,7 +24,7 @@
 
 ;;  structure here is {"prismatic/schem" ["0.1.1" "0.2.0" ...]}
 (defonce artifacts (atom (if (.exists (io/as-file artifacts-file))
-                           (->> artifacts-file slurp read-string (into (sorted-map)))
+                           (->> artifacts-file slurp edn/read-string (into (sorted-map)))
                            {})
                          :meta {:last-modified
                                 (get-last-modified-from-file artifacts-file)}))
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -10,6 +10,7 @@
             [refactor-nrepl.ns.slam.hound.regrow :as slamhound-regrow]
             [version-clj.core :as versions])
   (:import java.util.Date
+           java.util.zip.GZIPInputStream
            java.util.jar.JarFile))
 
 ;;  structure here is {"prismatic/schem" ["0.1.1" "0.2.0" ...]}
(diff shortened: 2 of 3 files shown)


paulrd:
This addresses #263

src/refactor_nrepl/artifacts.clj
```diff
@@ -10,10 +10,23 @@
             [refactor-nrepl.ns.slam.hound.regrow :as slamhound-regrow]
             [version-clj.core :as versions])
   (:import java.util.Date
+           java.util.zip.GZIPInputStream
            java.util.jar.JarFile))
 
+(def artifacts-file ".artifacts-cache")
```
expez: I think we should move this to the platform's temp dir:

1. That way we can share it across all projects, independently of the dir the REPL was started in
2. It makes it unnecessary to add this file to the `.gitignore` of every project you work on or your git config.

.gitignore
```diff
@@ -24,4 +24,5 @@ install.cmd
 /.cljs_node_repl/
 /out/
 /.lein-env
-.inline-deps
\ No newline at end of file
+.inline-deps
+/.artifacts-cache
```
expez: Should probably add a small note to the changelog too.

paulrd:
Cool I'll make those changes. I'm having a hard time testing this in emacs, how do you go testing this within emacs?

paulrd:
runnig ./build install i get the following:
Exception in thread "main" java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter, compiling:(cljs/util.cljc:1:1)

paulrd:
I think I know my issue with build.sh, I keep forgetting about the proper profile.

paulrd:
Testing in emacs is giving me different behavior. It's downloading on every invocation of jack-in. It is getting the gzipped version at least :-).

paulrd:
Is make inline-deps always required before I do lein install?

paulrd:
ah! :force true is set! why!?

paulrd:
Ah, ok it's not a bug. The default custom variable cljr-populate-artifact-cache-on-startup is true. Changing it to nil fixes the issue. I might open an issue there just as a reminder.

paulrd:
One more thing I should change I should either keep the map sorted or resort it when loading from cache file.

src/refactor_nrepl/artifacts.clj
```diff
@@ -10,10 +10,24 @@
...
+(defn get-last-modified-from-file
+  "Returns last modified time in milliseconds or nil if file does not exist."
+  [file]
+  (let [lm (.lastModified (io/file file))]
+    (if (zero? lm) nil lm)))
+
 ;;  structure here is {"prismatic/schem" ["0.1.1" "0.2.0" ...]}
-(defonce artifacts (atom {} :meta {:last-modified nil}))
+(defonce artifacts (atom (if (.exists (io/as-file artifacts-file))
+                           (->> artifacts-file slurp read-string (into (sorted-map)))
```
expez: We should be able to use `edn/read-string` as a drop-in replacement here and avoid exposing the full reader like this.  As you probably know `read-string` is capable of evaluating any code inserted into that file.

paulrd: Haha my spidy senses tingled on that but for some reason I never really understood that that's one of the things edn brought to the table. Thanks!

expez: You could also bind [\*read-eval\*](https://clojuredocs.org/clojure.core/*read-eval*) to `false`, but it reads way better to just use the EDN reader when possible :)

bbatsov:
Thanks!

Issue #267, opened 2019-10-04
Fallback on clojars when search.maven can't find artifact


yannvanhalewyn:
Search.maven can't find all the artifacts that are commonly used in clojure projects. Often these libraries can be found on Clojars. Examples of this are: [aero](https://clojars.org/aero), [clj-http](https://clojars.org/clj-http), [clj-time](https://clojars.org/clj-time), ... This results in `cljr-update-project-dependency` to throw the error `cljr--get-versions-from-middleware: Empty version list received from middleware!`. A secondary effect of this is that it stops walking all dependencies in a project when using `cljr-update-project-dependencies` when it hits one of these.

I'm happy to write a pull request for this. Before I do this I'd like to suggest what I'd like to change and see if this is a welcome contribution.

I would add a fallback to getting the version from clojars if non was returned from the search.maven API. Clojars has a straightforward API that returns artifact version. Examples:

https://clojars.org/api/artifacts/com.taoensso/timbre
https://clojars.org/api/artifacts/aero

It's pretty straightforward. Are there any problems I'm missing with this? Is this a welcome contribution?

Thanks for this great library I use actively every day!

yannvanhalewyn:
Closing this issue in favor of the pull request #268

PR #268, opened 2019-10-06
Fallback to Clojars for finding artifact version

yannvanhalewyn:
I had opened issue #267. This is the PR to solve the issue. This will use the Clojars API to fetch artifact versions if there are none to be found in Maven. The goal of this is to allow `cljr-update-project-dependency(ies)` to work with Clojars-only libraries, such as clj-http, clj-time and aero.


src/refactor_nrepl/artifacts.clj
```diff
@@ -101,6 +101,14 @@
                    (map #(vector (str group-id "/" %) nil))))
             group-ids)))
 
+(defn get-clojars-versions!
+  "Fetches all the versions of particular artifact from Clojars."
+  [for-artifact]
```
bbatsov: I'd name this just `artifact`.

src/refactor_nrepl/artifacts.clj
```diff
@@ -125,7 +133,8 @@
 (defn artifact-versions
   [{:keys [artifact]}]
   (->> (or (get @artifacts artifact)
-           (get-mvn-versions! artifact))
+           (seq (get-mvn-versions! artifact))
```
bbatsov: Might be a good idea to explain the behaviour of this function in the docstring.

src/refactor_nrepl/artifacts.clj
yannvanhalewyn: Me too, I followed the convention of the other function. Will update!

src/refactor_nrepl/artifacts.clj
yannvanhalewyn: 👍

bbatsov:
Thanks!

Issue #239, opened 2019-02-03
clean-ns: false positives for symbol usage


vemv:
Hi!

Sometimes I may want to `:require` something just for using the alias later for symbol construction, with no var usage.

For example:

```clojure
(ns om.g
  (:require
   [protocols.xyz :as xyz]))

(defn abc-impl [this] ...)

(defn new []
  ^{`xyz/abc abc-impl}
  {})
```

The above is a vanilla usage of metadata-based protocol extension, a new Clojure 1.10 feature.

If the ns didn't use `xyz` for any vars, `clean-ns` will wipe the require. And Clojure won't complain when compiling the ns again, since:

```
`xyz/abc
```

...is always a valid token.

Thanks - Victor

vemv:
Thanks!

That will help for the moment, had forgotten about the possibility.

Anyway I hope the root cause can get solved eventually.

I'd love to give it a shot myself, but can't at the moment. Busy life...

expez:
We don't look for symbols in the metadata, which is why this isn't handled.  

I'm not sure how much slower this would be if we also looked at the metadata of everything.  It might not be worth it to to handle this in general.  In other words, to avoid not making `clean-ns` too slow, we might have to introduce a new setting to also consider metadata.  At that point it might be just as easy to use the whitelist?

There's a half-finished PR in this repo about adding metadata to keep certain symbols from getting cleaned out.  So that would be another solution to this.

At work I have one or two namespace forms that are immediately followed by a `(require ...)` form with a comment about this ns getting loaded for side-effects and it's in an explicit require to avoid tooling removing it when cleaning the ns form.  If it was only me on the team I would've whitelisted it, but I'm not sure what the other tools would do so I left it like that.

Thanks for the report!

vemv:
Gotchu!

Just let me make sure - wouldn't it be pretty fast to just `slurp` a given .clj file and do a plain naive search of ocurrences of a given alias?

e.g. for `[clojure.string :as str]` we grep that .clj for the `str/` string (or regex). This includes symbols (this issue), function calls in metadata (#240), and possibly other edge cases.

Admittedly this wouldn't be as powerful as a proper code analysis, but I think it'd work, with few possible false possitives/negatives.

I'd say that it's better to err on the side of caution (don't strip requires because of a spurious match) than the opposite (strip requires because of a non-exhaustive search).

expez:
[...]

Issue #194, opened 2017-04-11
cljr-clean-ns when importing a deftype


aiba:
If namespace A has a deftype that is imported by namespace B, then namespace B needs to also require A, or the import could fail.  For example:

```clj
(ns test.a)

(deftype MyType [])
```

```clj
(ns test.b
  (:require test.a)
  (:import test.a.MyType))

(def x (MyType.))
```

If you run `cljr-clean-ns` on `test.b`, it will remove the `require` of `test.a`.  Without that require, the `(deftype MyType [])` is never evaluated, and loading `test.b` will produce a `ClassNotFoundException`.

Have other folks run into this / have a workaround?

One idea would be to add a rule to `cljr-clean-ns`: whenever `path.to.SomeClass` is imported, and there is a clojure namespace `path.to`, then also require `path.to`.

## Environment & Version information

```
clj-refactor 2.3.0-SNAPSHOT (package: 20170407.553), refactor-nrepl 2.3.0-SNAPSHOT
CIDER 0.15.0snapshot (package: 20170403.402)
Clojure 1.9.0-alpha15
Leiningen 2.7.1 on Java 1.8.0_112
GNU Emacs 25.2.1 (x86_64-apple-darwin16.4.0, Carbon Version 157 AppKit 1504.81) of 2017-03-09
OS X 10.12.4
```

expez:
Is this a problem even after loading the code into the repl, e.g. with <kbd>C-c C-k</kbd>?

I just find it hard to believe that no-one has run into this until now!

I suppose it makes some sense...  Even after writing clojure professionally for some time I've yet to use the `deftype` form. 

Thanks for reporting!

aiba:
> Is this a problem even after loading the code into the repl, e.g. with C-c C-k?

If you manually load `test.a` with `C-c C-k`, then yes, the deftype is evaluated and `test.a.MyType` is generated so the import would work without a require.  However, if you then did a `clojure.tools.namespace.repl/refresh`, it would reload the code in dependency order, and since nothing requires `test.a`, it would fail.

> I just find it hard to believe that no-one has run into this until now!

I believe it.  A common pattern with deftype (and other dynamically-generated java classes) is to provide a clojure function that wraps the class constructor.  So consumers would require/refer the clojure function that constructs the type, rather than importing class.

Another reason others might not have run into this is that often namespaces that have a deftype have other stuff in them which could cause that namespace to happen to be required before any of its generated classes are imported.

[...]

PR #270, opened 2019-11-04
Support inline whitelisting for :require/:imports used for side-effects

camsaul:
Tweaks `clean-ns` so that `:require` and `:import` symbols with `^:side-effects` metadata won't be removed. This is useful for cases where want import a namespace for side effects without having `clean-ns` remove it, or for false positives (such as #239).

###### before
```clj
(ns resources.ns-with-whitelisted-deps
  (:require [^:side-effects clojure.set :as set]
            [clojure.string :as str])
  (:import java.sql.Timestamp
           ^:side-effects java.util.Date))
```

###### after calling `clean-ns`

```clj
(ns resources.ns-with-whitelisted-deps
  (:require [clojure.set :as set])
  (:import java.util.Date))
```

Setting `cljr-libspec-whitelist`, the workaround mentioned in #239, doesn't work when running outside of Emacs – for example when using this [namespace declaration linter](https://github.com/camsaul/lein-check-namespace-decls) I put together using `clean-ns`.

To make this easier to use correctly it accepts `^:side-effects` on *any* of the following:

*  the `:require`/`:import` namespace/class symbol itself, e.g. 
    ```clj
    (:require ^:side-effects clojure.string)
    ```
*  the sequence wrapping the namespace/class symbol
    ```clj
    (:require ^:side-effects [clojure.string :as str])
    ```
*  the prefix form wrapping the namespace/class symbol
    ```clj
    (:require ^:side-effects [clojure data string])
    ```

### DIFF
--- a/src/refactor_nrepl/ns/ns_parser.clj
+++ b/src/refactor_nrepl/ns/ns_parser.clj
@@ -27,18 +27,17 @@
 (defn- expand-prefix-specs
   "Eliminate prefix lists."
   [libspecs]
-  (let [prepend-prefix (fn add-prefix [prefix libspec]
-                         (if (sequential? libspec)
-                           (apply vector
-                                  (symbol (str prefix "." (first libspec)))
-                                  (rest libspec))
-                           (symbol (str prefix "." libspec))))
-        normalize-libspec-vector (fn [libspec]
-                                   (if (core/prefix-form? libspec)
-                                     (let [prefix (first libspec)]
-                                       (map (partial prepend-prefix prefix)
-                                            (rest libspec)))
-                                     [libspec]))]
+  (letfn [(add-prefix [prefix libspec]
+            (if (sequential? libspec)
+              (into [(add-prefix prefix (first libspec))] (rest libspec))
+              (with-meta (symbol (str prefix "." libspec))
+                (merge (meta prefix) (meta libspec)))))
+          (normalize-libspec-vector [libspec]
+            (if (core/prefix-form? libspec)
+              (let [prefix (vary-meta (first libspec) (partial merge (meta libspec)))]
+                (map (partial add-prefix prefix)
+                     (rest libspec)))
+              [libspec]))]
     (mapcat normalize-libspec-vector libspecs)))
 
 (defn- use-to-refer-all [use-spec]
@@ -80,9 +79,10 @@
 (defn get-imports [ns-form]
   (let [expand-prefix-specs (fn [import-spec]
                               (if (sequential? import-spec)
-                                (let [package (first import-spec)]
+                                (let [package (vary-meta (first import-spec) (partial merge (meta import-spec)))]
                                   (map (fn [class-name]
-                                         (symbol (str package "." class-name)))
+                                         (with-meta (symbol (str package "." class-name))
+                                           (merge (meta package) (meta class-name))))
                                        (rest import-spec)))
                                 import-spec))]
     (some->> (core/get-ns-component ns-form :import)
--- a/src/refactor_nrepl/ns/prune_dependencies.clj
+++ b/src/refactor_nrepl/ns/prune_dependencies.clj
@@ -87,6 +87,7 @@
 (defn- class-in-use?
   [symbols-in-file c]
   (or
+   (:side-effects (meta c))
    ;; fully.qualified.Class
    (symbols-in-file c)
    ;; OnlyClassName, Class$Enum/Value or Class$Innerclass$InnerInnerClass
@@ -116,10 +117,11 @@
 ;; are side-effecting at load but might look unused and otherwise be
 ;; pruned.
 (defn- libspec-should-never-be-pruned? [libspec]
-  (let [ns-name (str (:ns libspec))]
-    (some (fn [^String pattern]
-            (re-find (re-pattern pattern) ns-name))
-          (:libspec-whitelist config/*config*))))
+  (or (:side-effects (meta (:ns libspec)))
+      (let [ns-name (str (:ns libspec))]
+        (some (fn [^String pattern]
+                (re-find (re-pattern pattern) ns-name))
+              (:libspec-whitelist config/*config*)))))
 
 (defn- prune-libspec [symbols-in-file current-ns libspec]
   (if (libspec-should-never-be-pruned? libspec)
--- a/test/refactor_nrepl/ns/clean_ns_test.clj
+++ b/test/refactor_nrepl/ns/clean_ns_test.clj
@@ -25,6 +25,8 @@
 (def ns-with-rename (clean-msg "test/resources/ns_with_rename.clj"))
 (def ns-with-rename-cleaned (core/read-ns-form-with-meta "test/resources/ns_with_rename_cleaned.clj"))
 (def ns-with-unused-deps (clean-msg "test/resources/unused_deps.clj"))
+(def ns-with-whitelisted-deps (clean-msg "test/resources/ns_with_whitelisted_deps.clj"))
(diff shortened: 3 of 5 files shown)


camsaul:
Let me know if you're happy with this approach, if so, I can add dox about it as well

aiba:
This could also be used as a workaround for #194

expez:
Hi, thanks for giving a damn and opening a PR! :100: 

We've had a similar PR open for discussion a while back and at that point I made the note that instead of `^:keep` we should namespace the keyword as `:refactor-nrepl/keep` so team members using other tooling had some way to google their what way to finding out what in the hell this metadata does.

In thinking about this now, though, I think this might be the wrong solution to the problem of side-effecting namespaces getting pruned from the ns.  My thinking is that we probably don't want to get into the business of adding tooling-specific metadata to source code.  On my team we have three different editors in use, can you imagine if each of these tools took the same approach?

```clj
(ns resources.ns-with-whitelisted-deps
  (:require [^:refactor-nrepl/side-effects ^:orchard/keep :^cursive/side-effecting ^:blessed  clojure.set :as set]
            [clojure.string :as str])
  (:import java.sql.Timestamp
           ^:refactor-nrepl/side-effects ^:orchard/keep :^cursive/side-effecting ^:blessed java.util.Date))
```

clj-condo recently had the same [discussion](https://github.com/borkdude/clj-kondo/issues/241) and landen on a solution they were happy with.  Unfortunately, that solution only applies to `require` statements and therefore only solves half the problem.

I think we should stick with solving this using the config options (which can be combined with dir-local settings so every emacs user on a project doesn't have to update their settings). With a slightly broader audience in mind, perhaps Orchard might at some point get a per-project config file that we can piggy-back on.

As to your particular problem in your project, @camsaul, you can configure the call to `clean-ns` by wrapping it with `(refactor-nrepl.config/with-config {:libspec-whitelist ["java.util.Date" ...]}...)`  More details about the available settings in the readme.

magnars:
How about not removing required namespaces with no alias and no referrals?

expez:
> How about not removing required namespaces with no alias and no referrals? 

This is how clj-condo solved it, but that leaves side-effecting imports in the lurch.

aiba:
Agree on not having tooling-specific metadata.  I also work on a team with different editors.

> How about not removing required namespaces with no alias and no referrals

This starts to reduce the utility of `cljr-clean-ns`.  Sometimes you really do want to require the whole namespace without aliasing it, for example a web server with 10 different `ring.middleware.foo` requires, each of which is only used once to write `ring.middleware.foo/wrap-foo`.  I'd rather not have to add 10 different aliases in this case.  And yet if I remove a call to the middleware, I do want `cljr-clean-ns` to remove the require, since it has no side effects and is not used anymore.

Other times, the required namespace consists of only a root name, for example [timbre-ns-pattern-level](https://github.com/yonatane/timbre-ns-pattern-level).


All the proposed solutions on this thread have some kind of inelegance, and I would suggest that they be weighed against the inelegance of existing workarounds.  My current workaround is use the side-effecting namespace somehow, for example:

```
(ns my-ns
  (:require [something.with.side.effects]))

something.with.side.effects/some-var ;; ns has side-effects
```

It only adds one line to the code, makes it pretty clear that there are side effects, and should work across editors.

One challenge I run into is if the side-effecting namespace has no vars, for example [tick.timezone](https://github.com/juxt/tick/blob/master/src/tick/timezone.cljc).  For this I resort to an uglier workaround:

```
(ns my-ns
  (:require [tick.timezone]))

(comment
  tick.timezone/dont-clean-this-require
  )
```

But I could imagine a future where this no longer works because ns usages inside comments no longer prevent cleaning.

So my question for this discussion is: rather than metadata, is there a simple form that can be placed at the top level which involves a required namespace to prevent its cleaning?  And which works even if the required ns has no vars in it?  If there is such a form, then maybe that's a reasonable solution to this problem.

expez:
> If there is such a form, then maybe that's a reasonable solution to this problem.

The language already has such a form, thankfully.  

Lately I've just been putting an explicit `(import ...)` or `(require ...)`, with a comment about the desired side-effects, on its own line below the `ns` form.  e.,g

```clj
(ns my-ns
  (:require [clojure.string :as str]))

;; This has some really sweet side-effects that we want, because reasons
(require 'tick.timezone) 
```

This is something I've started doing to help the others on my team (even though I have the right variables set up to avoid accidental pruning), and it has a few benefits:

1. It clearly shows that this `import` / `require` is special and requires some extra attention.
2. Easily understandable to anyone
3. Doesn't require any out-of-band communication (e.g. you don't have to know about tooling metadata or conventions with additional meanings like `[this-will-be.left-alone]`)
4. Looks better than adding a comment in the `ns` form anyway, that may or may not be removed by tooling
5. Robust (we've lost important side-effecting imports  to tooling bugs / eager janitors).

magnars:
That’s a really neat solution. 👍

ons. 6. nov. 2019 kl. 14:16 skrev Lars Andersen <notifications@github.com>:

aiba:
> Lately I've just been putting an explicit (import ...) or (require ...), with a comment about the desired side-effects, on its own line below the ns form. e.,g

I love it!  I would be happy to write all my code this way and continue to use `cljr-clean-ns` as it is.

But does it work in clojurescript?  The [docs](https://cljs.github.io/api/cljs.core/require) say "Only usable from a REPL".

camsaul:
My strategy in the past has been to add a separate top-level `(require ...)` for things that would normally get removed. But ultimately it's preferable to put everything in the top-level `ns` form because a few tools out there use it for things like building dependency graphs between namespaces.

[`ns-tracker`](https://github.com/weavejester/ns-tracker), which is used by Ring for reloading, for example, only looks at the `ns` form. So by using `require` instead you could maybe get your web server in a weird state when developing locally.

Another example is one of Eastwood's linters which [checks for referencing vars without requiring their namespace first](https://github.com/jonase/eastwood#implicit-dependencies). This linter also only looks at the `ns` declaration and not at top-level `require` or `import` forms.

I'll have to look at either using a combo of a global config and `.dir-locals.el` or something hacky like wrapping fake references in `(comment ...)` so I can have my cake and eat it too

aiba:
@camsaul thanks for pointing this out.  In my case, I use [tools.namespace](https://github.com/clojure/tools.namespace), which I just checked and also only looks at the top-level `ns` form to determine dependencies.

Well it's a kludge, but for now I think I'm going to write:

```
(ns my-ns
  (:require [tick.timezone]))

(comment tick.timezone/side-effects)
```

I wish there were a better form to put in the body, but I can't think of one.

bbatsov:
So, how are we going to proceed here? I don't quite remember what kondo opted for, but I guess we should take the same approach here.

expez:
> So, how are we going to proceed here?

I'm in favor of just closing this because there's no good way solving this.  

In summary: the practice I suggested above works really well for clj, but not at all for cljs.  The clj-condo approach doesn't work for imports, and also depends on educating users that they can get a certain outcome if they adopt a certain code style.

olymk2:
I am hitting this as well, in particular with specs I require files which contain specs so that they are loaded into the global registry but running clean will remove the require and break the code.

going to use the comment hack for now.

expez:
Closing this in favor of [this issue](https://github.com/clojure-emacs/refactor-nrepl/issues/292) and enabling opt-in support for the clj-kondo approach behind a setting.

clj-kondo is great and it's usage has been increasing steadily.  Over time I expect it to be the dominant linter (if it isn't already).  At that point the argument about the need to educate users goes out the window (since they do that for us and have a much larger user base).  If we make it opt-in it would only affect users that want this behavior.

PR #272, opened 2019-11-26
Preserve npm string-style requires for cljr-clean-ns

aiba:
Fixes clojure-emacs/clj-refactor.el#453

### DIFF
--- a/src/refactor_nrepl/ns/rebuild.clj
+++ b/src/refactor_nrepl/ns/rebuild.clj
@@ -97,16 +97,22 @@
     :all
     (sort dependency-comparator referred)))
 
+(defn- type-preserving [f x]
+  (when x
+    (when-let [r (f x)]
+      (cond
+        (string? x) (str r)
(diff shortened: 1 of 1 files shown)


aiba:
In the course of creating this PR, I came across the function `create-prefixed-libspec-vector` which is awfully similar to `create-libspec-vectors-without-prefix`, and strangely it calls `(assoc libspec :ns (ns-suffix ns))`, but `ns-suffix` takes a libspec, not a ns, so I wonder if this is a separate bug.

I don't think this should hold up the merging of this PR to address the bug that this PR fixes, but just an obversation.

expez:
Thanks for this!

> strangely it calls `(assoc libspec :ns (ns-suffix ns))`, but `ns-suffix` takes a libspec, not a ns, so I wonder if this is a separate bug.

Definitely sounds like it!

expez:
Unfortunately there's a problem with the deploy script, so I'm not sure when the next snapshot will go out.

I might have some time to investigate next week.

PR #273, opened 2019-11-27
remove unused function

aiba:
Per #272, this function has a bug in it and I checked and it's actually never called, so let's just get remove it :-).


bbatsov:
Fine by me. Thanks!

PR #274, opened 2019-11-27
clean-ns: group together npm deps

aiba:
This small change will have all the npm string-style requires grouped together.  I think this is desirable, as npm libraries look and act quite differently from regular clojure requires.

### DIFF
--- a/src/refactor_nrepl/ns/rebuild.clj
+++ b/src/refactor_nrepl/ns/rebuild.clj
@@ -60,7 +60,7 @@
   ^String [dep]
   (str/lower-case
    (if (sequential? dep)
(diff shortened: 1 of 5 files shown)


expez:
Sounds like a good idea!

Would be nice to have a test for this so the functionality doesn't break when people work on the code base.

aiba:
I added some tests.

I had to also add a workaround in `refactor-nrepl.ns.tracker` for [this clojure.tools.namespace bug](https://clojure.atlassian.net/projects/TNS/issues/TNS-51).

expez:
Terrific!  Thanks!

PR #276, opened 2019-12-08
[Fix #265] Update orchard version

### LINKED ISSUE #265: Warning of illegal reflective access on Java 12
refactor-nrepl currently uses orchard 0.4.0, which bundles old dynapath, which causes Java 12 to spit following warning:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by mranderson048.orchard.v0v3v0.dynapath.v0v2v5.dynapath.defaults$eval28489$fn__28490 to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of mranderson048.orchard.v0v3v0.dynapath.v0v2v5.dynapath.defaults$eval28489$fn__28490
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

I wonder if it's possible to update to a newer orchard, to remove this warning? :) I guess it's impossible to do in my project.clj because this dependency is inlined, right?

mbuczko:
Unfortunately the change was not that straightforward due to how orchard 0.5.4 calculates classpaths and refactored the functions around.

I hope I adjusted the code here well enough to follow all the classpath-related changes in orchard. Let me know if there is something wrong or missing.

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -3,7 +3,8 @@
             [clojure.string :as str]
             [clojure.tools.namespace.parse :as parse]
             [clojure.tools.reader.reader-types :as readers]
-            [orchard.classpath]
+            [orchard.java.classpath :as cp]
+            [orchard.misc :as misc]
             [me.raynes.fs :as fs]
             [refactor-nrepl.util :refer [normalize-to-unix-path]]
             [refactor-nrepl.s-expressions :as sexp]
@@ -46,8 +47,11 @@
   directory and paths matching :ignore-paths specified in config.
   Follows the semantics of orchard classpath."
   []
-  (->> (orchard.classpath/classpath-directories)
-       (remove #(-> % str normalize-to-unix-path (.endsWith "target/srcdeps")))))
+  (->> (cp/classpath)
+       (remove misc/archive?)
+       (keep #(let [f (io/file %)]
+                (when (.isDirectory ^File f) f)))
+       (remove (comp ignore-dir-on-classpath? str))))
 
 (defn project-root
   "Return the project root directory.
@@ -95,7 +99,6 @@
   are pruned by this function."
   [pred dir]
   (->>  dir
-        io/file
         file-seq
         (filter (every-pred fs/exists?
                             (complement fs/hidden?)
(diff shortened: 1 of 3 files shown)


src/refactor_nrepl/core.clj
```diff
@@ -95,7 +100,6 @@
   are pruned by this function."
   [pred dir]
   (->>  dir
-        io/file
```
mbuczko: seems like it's not needed. this function is used with input coming from `dirs-on-classpath` which already returns `java.io.File`s.

src/refactor_nrepl/core.clj
```diff
@@ -46,8 +47,12 @@
   directory and paths matching :ignore-paths specified in config.
   Follows the semantics of orchard classpath."
   []
-  (->> (orchard.classpath/classpath-directories)
-       (remove #(-> % str normalize-to-unix-path (.endsWith "target/srcdeps")))))
+  (->> (cp/classpath)
+       (filter (complement misc/archive?))
```
expez: Could use `remove` here to avoid the `complement`.

src/refactor_nrepl/core.clj
```diff
@@ -46,8 +47,12 @@
...
+       (map #(let [f (io/file %)]
```
expez: If you use `keep` here you can get rid of the `nil?` check on the line below.

expez:
Other than some minor nitpicks, this looks good to me @mbuczko.  Thanks for taking this on!

The commit message could also do with some capitalization and a change to the present tense, to be in the line with the rest of the git history.

bbatsov:
Looks good to me. Not sure why we removed a couple of functions from the API, a bit of time passed and my memory is fuzzy. Nicely done!

src/refactor_nrepl/core.clj
```diff
@@ -46,8 +47,11 @@
   directory and paths matching :ignore-paths specified in config.
   Follows the semantics of orchard classpath."
   []
-  (->> (orchard.classpath/classpath-directories)
-       (remove #(-> % str normalize-to-unix-path (.endsWith "target/srcdeps")))))
+  (->> (cp/classpath)
+       (remove misc/archive?)
```
mbuczko: @expez `filter` changed to `remove` as suggested. thanks for review.

expez:
Thanks @mbuczko! 👍

PR #277, opened 2019-12-13
[Fix #275] Honor cljr-ignore-analyzer-errors flag in macros analyzer

### LINKED ISSUE #275: Invalid Keyword in unrelated file/namespace when running `find-symbol`
## Expected behavior
`find-symbol` returns symbol usages within the project boundaries

## Actual behavior
`find-symbol` breaks when trying to find symbol usages on symlinked dependencies' test files.

To give more context than one sentence:
I work on a monorepo project and we use `deps.edn` to manage our dependencies. One of those dependencies is a lib of common code within the same monorepo, which we symlink to the project in a `deps/` folder and reference it using `lib/common {:local/root "deps/lib.common"}`.
We have to do it this way because the build script (nix) will consider the project directory as a root directory, so we can't have references to `../../libs/lib.common`.

It finds tries to find symbols in every file in the folder, but as it goes through the symlink, it reaches `test/**/*.clj` on this lib, which dependencies we might not have or might not have loaded on the current file.

It then breaks with the old `::ns/kw` error of `ns` not being known and the `edn-reader` breaks.

Tracking specifically this case, it broke on `find-macros`, `find-macro-definitions-in-file`.

## Steps to reproduce the problem

Self-evident from previous explanation, but I'll summarize:

1 - Create two top level projects
2 - Symlink first on `second/deps/first`
3 - Reference first on second's `deps.edn`
4 - Create a test file on first, with: `(require [some.ns.from.first :as first])` and `::first/some-kw`
5 - Run `find-symbol` on any symbol on first (it can be anything, even the most unrelated symbol)

* It should break on the test namespace on `second/deps/first/test/test_file_test.clj`

## Environment & Version information

Using [acid.nvim](https://github.com/clojure-vim/acid.nvim):
```
cider-nrepl: 0.22.0
refactor-nrepl: 2.4.0
```

### Operating system

Fedora 31

mbuczko:
I feel like we should have a way to ignore macro analyzer exceptions the same way we do it in `find-global-symbol` here https://github.com/clojure-emacs/refactor-nrepl/blob/75f5255f0e4e701f50d30cc58216e03c2d558905/src/refactor_nrepl/find/find_symbol.clj#L133.

Global symbols / macros analyzing can be really tricky sometimes and it fails with similar errors in some of my own projects too. Unfortunately, setting `cljr-ignore-analyzer-errors flag` is not taken into consideration in macro search process. I tried to improve it here.

### DIFF
--- a/src/refactor_nrepl/find/find_macros.clj
+++ b/src/refactor_nrepl/find/find_macros.clj
@@ -51,6 +51,7 @@
   (util/with-additional-ex-data [:file (.getAbsolutePath f)]
     (with-open [file-rdr (FileReader. f)]
       (binding [*ns* (or (core/path->namespace :no-error f) *ns*)
+                reader/*read-eval* false
                 reader/*data-readers* *data-readers*]
         (let [rdr (LineNumberingPushbackReader. file-rdr)
               opts {:read-cond :allow :features #{:clj} :eof :eof}]
@@ -81,9 +82,13 @@
 
 (defn- find-macro-definitions-in-project
   "Finds all macros that are defined in the project."
-  []
+  [ignore-errors?]
   (->> (core/find-in-project (some-fn core/cljc-file? core/clj-file?))
-       (mapcat get-macro-definitions-in-file-with-caching)))
+       (mapcat #(try
+                  (get-macro-definitions-in-file-with-caching %)
+                  (catch Exception e
+                    (when-not ignore-errors?
+                      (throw e)))))))
 
 (defn- get-ns-aliases
   "Create a map of ns-aliases to namespaces."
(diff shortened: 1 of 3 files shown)


src/refactor_nrepl/find/find_macros.clj
```diff
@@ -51,6 +51,7 @@
   (util/with-additional-ex-data [:file (.getAbsolutePath f)]
     (with-open [file-rdr (FileReader. f)]
       (binding [*ns* (or (core/path->namespace :no-error f) *ns*)
+                reader/*read-eval* false
```
mbuczko: I'm not sure about this one. Do we really need to evaluate analyzed expressions?

expez: > What would be the disadvantage of switching evaluation off?

Taking time out of your life to figure out if it's the correct thing to do or not 🤣 

Don't think this will make any difference in this case.

expez: > What would be the disadvantage of switching evaluation off?

Taking time out of your life to figure out if this is correct or not? 🤣 

I don't think this will make any difference in this situation.  People never use the reader to eval anyway, so even if this is the wrong thing to do nobody will be affected.

PR #280, opened 2019-12-14
[clj-refactor.el #443] Add support for js literal

Frozenlock:
As described in
https://github.com/clojure-emacs/clj-refactor.el/issues/443, it was
impossible to use clj-clean-ns when a js literal was present:

'No reader function for tag js'

One possible solution was to use *default-data-reader-fn* with the
function `(fn [_tag value] value)`. However this had the potential of
swallowing useful error messages when dealing with other undefined
tags.

The chosen solution is to provide a dummy reader function specifically
for the #js tag (and only for the :cljs dialect). Rather than
transforming the provided form into a JS object, it returns the
clojure form, which can then be analyzed normally.

---------

- [N/A] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [X] You've added tests (if possible) to cover your change(s)
- [X] All tests are passing (run `lein do clean, test`)
- [X] Code inlining with mranderson works and tests pass with inlined code (run `./build.sh install` -- takes a long time)
- [X] You've updated the changelog (if adding/changing user-visible functionality)
- [N/A] You've updated the readme (if adding/changing user-visible functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* [clojure-emacs/clj-refactor.el#443](https://github.com/clojure-emacs/clj-refactor.el/issues/443) `clean-ns` support namespaces with js literal (#js).
 * [#251](https://github.com/clojure-emacs/refactor-nrepl/pull/251) `clean-ns` support extra message key `relative-path`, which will be used if `path` does not exist.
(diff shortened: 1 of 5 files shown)


expez:
Excellent!  Thanks @Frozenlock! 👍

PR #294, opened 2021-02-18
Treat uneval nodes as comments

tobias:
Uneval nodes throw an UnsupportedOperationException if node/sexpr is
called on them, which will break get-first-sexp if it is given
file-content that starts with `#_ whatever`.

This includes uneval nodes in the set of nodes we should skip when
looking for first & last sexps.

- [X] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
  - there is no `CONTRIBUTING.md`, I assume this should be `.github/CONTRIBUTING.md`?
- [X] You've added tests (if possible) to cover your change(s)
  - No! I get:
```
Exception in thread "main" java.lang.RuntimeException: Map literal must contain an even number of forms, compiling:(refactor_nrepl/artifacts.clj:25:1)
```
But I get the same thing on `master`, so I'm not sure what is going on there.
 
  - There is no `./build.sh`. I have installed this with `lein with-profile +plugin.mranderson/config install` and have confirmed it is working for me.
- [X] You've updated the changelog (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* [#294](https://github.com/clojure-emacs/refactor-nrepl/pull/294): Properly skip uneval nodes when looking for the first/last sexp
+
 ## 2.5.1 (2021-02-16)
 
 ### Bugs fixed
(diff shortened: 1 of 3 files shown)


bbatsov:
> But I get the same thing on master, so I'm not sure what is going on there.

Well, it I see on the CI tests mostly pass (apart from the known issue with Clojure 1.8). Thanks for fixing this!

Issue #296, opened 2021-04-02
clean-ns rebuilds namespaces incorrectly if they only consist of conditionals for one language


camsaul:
`refactor-nrepl.ns.clean-ns/clean-ns` with a `.cljc` namespace like

```clj
(ns conditional-with-require-macros
  #?(:cljs (:require-macros [conditional-with-require-macros :refer [my-when]])))
```

returns

```clj
(ns conditional-with-require-macros
  (:require-macros [conditional-with-require-macros :refer [my-when]]))
```

It incorrectly strips out the conditionals. However, `refactor-nrepl.ns.clean-ns/clean-ns` with a namespace declaration that has some shared `:requires` like

```clj
(ns conditional-with-require-macros
  (:require [clojure.string :as str])
  #?(:cljs (:require-macros [conditional-with-require-macros :refer [my-when]])))
```

returns

```clj
(ns conditional-with-require-macros
  (#?@ (:clj [(:require [clojure.string :as str])]
        :cljs [(:require [clojure.string :as str])
               (:require-macros [conditional-with-require-macros :refer [my-when]])])))
```

preserving the conditionals.

I've narrowed it down and it's a bug with `rebuild/build-cljc-dep-forms` specifically -- happening because of the check at https://github.com/clojure-emacs/refactor-nrepl/blob/master/src/refactor_nrepl/ns/rebuild.clj#L248-L250.

Incorrect result:

```clj
(refactor-nrepl.ns.rebuild/build-cljc-dep-forms
 '{:clj            {:require ()
                    :import  ()}
   :cljs           {:require        ()
                    :import         ()
                    :require-macros ({:ns conditional-with-require-macros, :refer (my-when), :rename {}})}
   :source-dialect :cljc})

->

((:require-macros [conditional-with-require-macros :refer [my-when]]))
```

Correct result:

```clj
(refactor-nrepl.ns.rebuild/build-cljc-dep-forms
 '{:clj            {:require ({:ns clojure.string, :as str, :rename {}})
                    :import  ()}
   :cljs           {:require        ({:ns clojure.string, :as str, :rename {}})
                    :import         ()
                    :require-macros ({:ns conditional-with-require-macros, :refer (my-when), :rename {}})},
   :source-dialect :cljc})

->

(#?@ (:clj [(:require [clojure.string :as str])]
      :cljs [(:require [clojure.string :as str])
             (:require-macros [conditional-with-require-macros :refer [my-when]])]))
```

PR #297, opened 2021-04-02
Fix build-cljc-dep-forms with conditionals for only one language 

camsaul:
Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/296


expez:
Thank you so much, @camsaul! 💯

PR #298, opened 2021-04-13
Fix the build / honor the `refresh-dirs`

vemv:
Fixes the build with the following strategy:

* isolate the "Lein plugin" part to its own source path
  * this makes it possible to exercise older versions of Clojure, which don't have e.g. `*print-namespace-map*`, which the Lein implementation can assume.
  * This is also the strategy followed in Eastwood (to name another project that is both a library and a Lein plugin at the same time)
* When tools.namespace `refresh-dirs` are set, don't analyze files outside those refresh dirs.
  * This is a quality of life improvement anyway. A properly set up Clojure project (that is using tools.namespace) tends to have the `refresh-dirs` set.
  * This way end users don't have extraneous code being possibly analyzed: examples, drafts, production scripts, etc all of which are dangerous to analyze (because analysis == execution of top-level forms).
  * For this to work, tools.namespace is resolved in runtime.
    * `cider-refresh` does the same: https://git.io/JOYUT


test/global_test_setup.clj
```diff
@@ -0,0 +1,8 @@
+(ns global-test-setup
+  "This namespace's contents will be automatically loaded by `lein test` (or any test runner)."
+  (:require
+   [refactor-nrepl.inlined-deps.toolsnamespace.v1v1v0.clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
```
vemv, line 4: Ideally this would be:

```diff
- refactor-nrepl.inlined-deps.toolsnamespace.v1v1v0.clojure.tools.namespace.repl
+ clojure.tools.namespace.repl
```

but I couldn't wait get it to work. It looked like mranderson :expositions would do the trick but nope?

src/refactor_nrepl/ns/tracker.clj
```diff
@@ -37,8 +55,28 @@
           :when ((set deps) ns)]
       file)))
 
+(defn- in-refresh-dirs? [refresh-dirs file]
```
vemv, line 58: Can add a unit test to this one. For now I didn't b/c my reasoning was that this is a private helper, and often private helpers are left untested.

Don't know how `clojure-emacs` folks tend to approach these - would be happy to tweak

test/global_test_setup.clj
bbatsov: It seems like this old issue to me https://github.com/benedekfazekas/mranderson/issues/5

src/refactor_nrepl/ns/tracker.clj
bbatsov: I don't typically add tests for private helpers.

bbatsov:
The changes look good to me. Thanks!

PR #299, opened 2021-04-14
Address most reflection warnings / cleanup CI

vemv:
* Run clj-kondo by a more stable means
* Run Eastwood in CI again
  * only src for now, as the `test` setup is a bit intrincate with multiple/nested source-paths
* CI: adopt matrix syntax
* Address most reflection warnings
  * All `src/` warnings are gone


bbatsov:
Great work!

PR #300, opened 2021-06-24
Misc fixes

expez:
Could I trouble you to merge this and cut a snapshot release, @bbatsov?  

Currently on a Windows machine and mranderson doesn't work here so I can't install locally or push to clojars.  Haven't had time to investigate why, but would like to dogfood these changes prior to a new point release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+* [clojure-emacs/clj-refactor.el#459](https://github.com/clojure-emacs/clj-refactor.el/issues/459): `clean-ns` should conform to the style guide: `(:require` in the ns form should be followed by a newline.
 * [#294](https://github.com/clojure-emacs/refactor-nrepl/pull/294): Properly skip uneval nodes when looking for the first/last sexp
 * From now on, if you set the `clojure.tools.namespace.repl/refresh-dirs`, files outside said `refresh-dirs` won't be analyzed, resulting in safer, more efficient analysis.
+
 
 ## 2.5.1 (2021-02-16)
 
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject refactor-nrepl "2.5.1"
+(defproject refactor-nrepl "3.0.0-SNAPSHOT"
   :description "nREPL middleware to support editor-agnostic refactoring"
   :url "http://github.com/clojure-emacs/refactor-nrepl"
   :license {:name "Eclipse Public License"
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -19,7 +19,7 @@
 
 ;;  structure here is {"prismatic/schem" ["0.1.1" "0.2.0" ...]}
 (defonce artifacts (atom (if (.exists (io/as-file artifacts-file))
-                           (->> artifacts-file slurp edn/read-string (into (sorted-map)))
+                           (->> artifacts-file slurp edn/read-string (into {}))
                            {})
                          :meta {:last-modified
                                 (get-last-modified-from-file artifacts-file)}))
@@ -107,7 +107,7 @@
 (defn- get-artifacts-from-clojars!
   []
   (reduce #(update %1 (str (first %2)) conj (second %2))
-          (sorted-map)
+          {}
           (get-clojars-artifacts!)))
 
 (defn- update-artifact-cache!
--- a/src/refactor_nrepl/ns/pprint.clj
+++ b/src/refactor_nrepl/ns/pprint.clj
@@ -36,7 +36,7 @@
 
 (defn pprint-require-form
   [[_ & libspecs]]
-  (print "(:require ")
+  (print "(:require\n")
   (dorun
    (map-indexed
     (fn [idx libspec]
--- a/test/resources/artifacts_pprinted
+++ b/test/resources/artifacts_pprinted
@@ -1,13 +1,14 @@
 (ns refactor-nrepl.artifacts
-  (:require [clojure
-             [edn :as edn]
-             [string :as str]]
-            [clojure.data.json :as json]
-            [clojure.java.io :as io]
-            [nrepl
-             [middleware :refer [set-descriptor!]]
-             [misc :refer [response-for]]
-             [transport :as transport]]
-            [org.httpkit.client :as http]
-            [refactor-nrepl.externs :refer [add-dependencies]])
+  (:require
+   [clojure
+    [edn :as edn]
+    [string :as str]]
+   [clojure.data.json :as json]
+   [clojure.java.io :as io]
+   [nrepl
+    [middleware :refer [set-descriptor!]]
+    [misc :refer [response-for]]
+    [transport :as transport]]
+   [org.httpkit.client :as http]
+   [refactor-nrepl.externs :refer [add-dependencies]])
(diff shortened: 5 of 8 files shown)


vemv (commented):
While I like and regularly use the newly proposed newline style, a significant % of people don't, so I'd suggest keeping the traditional behavior and making the new one opt-in.

I know of teams that use refactor-nrepl as their official 'formatter' (that is, for ns forms).

WDYT?

expez:
> a significant % of people don't

Do people write the other way around on purpose when they are manually managing the ns form?

I believe that this way of formatting dominates the other one completely, because you greatly reduce the risk of having the ns form take up too much horizontal space.  Does the other wa of formatting have any redeeming qualities beyond being familiar?

> I'd suggest keeping the traditional behavior and making the new one opt-in

I *really* want the default behavior of the pretty printer to be in line with the [style guide](https://github.com/bbatsov/clojure-style-guide#line-break-ns-declaration).  I'd be way more open to adding a setting to explicitly preserve the old behavior.

While I have someone on the line that gives a shit about ns formatting, another thing that's been bugging me about the current pretty printer is this:

```clj
(:require
 [this.very.long.namespace.for.gets.printed.very.weirdly
  :as
  long-ns]
```

I believe this would be better as:

```clj
(:require
 [this.very.long.namespace.for.gets.printed.very.weirdly
  :as long-ns]
```

but didn't have the time to fix this today.  This happens because we delegate the printing of the vector to `clojure.pprint/pprint` and it will print long vectors like that.  If you had a long `:refer` it would do the same.  This isn't something I noticed before it's too late, but it's making the ns form take up too much vertical space.  Do you think this will also be contentious?

I have some other stuff I want to fix (e.g. hotload dependency) so maybe we can batch up the contentious stuff and do a 3.0.0 release where some breakage can be permitted.

vemv:
> Does the other way of formatting have any redeeming qualities beyond being familiar?

I'd be inclined to say that judging the merits of each approach is besides the point. Instead one can simply ohserve that the current approach has been there for many years, and also aligns with the way that many people have written Clojure historically.

I'd concur in that there's no big reason or advantage for the traditional style. It's just familiar, and saves one <kbd>RET</kbd> when typing it out. Aside that by now it's hardwired into the brains of a significant % of developers.

This is to say: not introducing breaking changes is valuable. RH's talks have increased the sensibility in that direction, so ideally we wouldn't disappoint people. Especially when refactor-nrepl has gradually some traction over the years; I'd say that for re-gaining it we should be careful with our users.

Finally, CIDER actually has some soft commitment to not further break changes, see e.g. https://github.com/clojure-emacs/cider/discussions/2960

> I really want the default behavior of the pretty printer to be in line with the style guide

Worth pointing out that the styleguide is actually worded quite leniently for this specific point, surely in acknowledgement that this was a contentious point.

vemv:
> I have some other stuff I want to fix (e.g. hotload dependency) 

Just in case, I assume you're familiar with https://github.com/clojure-emacs/orchard/issues/103 and related issues/PRs?

The whole goal appeared to be tricky - it could not work at all, or cause multiple bugs that were hard to track down, etc.

expez:
> This is to say: not introducing breaking changes is valuable. RH's talks have increased the sensibility in that direction, so ideally we wouldn't disappoint people. Especially when refactor-nrepl has gradually some traction over the years; I'd say that for re-gaining it we should be careful with our users.

Definitely agree with the sentiment, but that conversation was in the context of libraries / APIs.  I don't think we'll break any existing programs with this change, though we might surprise people a little.

The `clean-ns` functionality as a whole isn't perfect, but the point of it is to make something that's good enough that we never have to touch the ns form manually.  Manually formatting and editing the ns form is a very low value activity that we want to minimize.  My view is that in exchange for avoiding tedious our users have been willing to put up with certain bugs / idiosyncrasies, but would be even happier if we could resolve those issues.  This newline is one, the splitting of vectors across too many lines is another and then there are various non-contentious bugs e.g. with deftypes.

Along this line of reasoning comes the argument that would shouldn't put the most sensible defaults behind opt-in settings, but instead allow ourselves some breakage.

I'm really glad you chimed in.  I still want to do this, but I can definitely see the value in doing it in a 3.0.0 release and adding an explicit changelog entry.  Do you think that's enough @vemv or should I also add a new setting so people can get the old behavior back?

magnars:
It might not be the most pertinent point, but I manually "correct" these newlines every single time. I'm quite familiar with Emacs' M-j (join-line) after these years.

expez:
> It might not be the most pertinent point, but I manually "correct" these newlines every single time. I'm quite familiar with Emacs' M-j (join-line) after these years.

The humanity! 😱

expez:
Aight, bumped the next release to 3.0.0 and added a proper changelog entry for this 'fix' :)

vemv:
> I'm really glad you chimed in. I still want to do this, but I can definitely see the value in doing it in a 3.0.0 release and adding an explicit changelog entry. Do you think that's enough @vemv or should I also add a new setting so people can get the old behavior back?

Thanks! 3.0.0 is nice - often other projects follow semver more laxly.

Opt-out seems ok, would be the second best choice imo 😄 a Java system property tends to be nice for these cases - this way one can set it in a team's project.clj, removing the need to worry about individuals' emacs setups.

Would also be nice to keep the test coverage for the old formatting. It might be easy (without duplicating files such as `test/resources/artifacts_pprinted `) by using `string/replace` within the relevant `deftest`.

bbatsov:
I'll take a closer look at the proposed changes tomorrow/over the weekend.

bbatsov:
While I welcome the suggested change to the ns formatting, I'm thinking it'd be nice if we added some configuration flag that controls on the client-side whether there's a newline or not. That's trivial to do (just a `defcustom` and an extra param to the nREPL op that tracks it) and I guess it will keep happy the people who want to stick to the old behavior after the next release is out.

expez:
> While I welcome the suggested change to the ns formatting, I'm thinking it'd be nice if we added some configuration flag that controls on the client-side whether there's a newline or not. That's trivial to do (just a defcustom and an extra param to the nREPL op that tracks it) and I guess it will keep happy the people who want to stick to the old behavior after the next release is out.

Anyone willing to take that on?  If not I'm inclined to just drop the commit.  There are many other bugs in this project that I'd rather spend an hour on fixing.

bbatsov:
I can do this down the road myself, but I've been crazy busy at work for the past few months. I guess we can just open a ticket about this and hope that someone will get to doing it.

bbatsov:
> I'm really glad you chimed in. I still want to do this, but I can definitely see the value in doing it in a 3.0.0 release and adding an explicit changelog entry. Do you think that's enough @vemv or should I also add a new setting so people can get the old behavior back?

I should have also read the discussion so far - seems you already discussed this. We can also "cheat" a bit and use a Java property to control this. Anyways, not a big deal for me.

vemv:
No issue 👍

I'll add the toggle in a PR. I happen to have the time and was planning to contribute to refactor-nrepl!

bbatsov:
I've deployed a new snapshot.

bbatsov:
> I'll add the toggle in a PR. I happen to have the time and was planning to contribute to refactor-nrepl!

@vemv Great! I just noticed you can also add this as part of https://github.com/clojure-emacs/refactor-nrepl#configuration

I never used `clj-refactor`, I so I hadn't noticed it has one configuration map for everything. This makes the task even simpler than I thought it was.

PR #301, opened 2021-06-25
Bring hotload-dependency back 

expez:
Previously this op was built on top of alembic, but that lib no longer
seems to be maintained at all so we're moving to pomegranate.

`tools.deps` would've been another alternative by their `add-lib`
operation has yet to be merged to the main branch.  Once it's merged and
has proven itself in the wild nothing prevents us from moving over,
though, if that solution proves to be better in some way.


expez:
@benedekfazekas this seems to fail due to the `mranderson` pass.  Is that something you have time to investigate?

bbatsov:
Now that the other PR is merged, I guess you can rebase this one on top of `master`.

vemv:
I'd try to replicate this test as part of this PR: https://github.com/clojure-emacs/orchard/blob/dbe1bef72697a89843763dd112d151a9f365e5c8/test/orchard/java/classpath_test/third_party_compat_test.clj - else I'd fear that users of this feature could have issues that could manifest themselves in third-party libs.

(in this case the `when-not` conditional is not needed)

expez:
> Now that the other PR is merged, I guess you can rebase this one on top of master.

Done

> I'd try to replicate this test as part of this PR

Thanks.  That's a good idea.  I read some of the discussion you linked yesterday, but it's not entirely clear what this test is testing.

```clj
(deftest works
    (is (seq (clojure.java.classpath/classpath-directories))
        "The presence of `clojure.java` does not affect third-party libraries"))
```

If you get stuff back it works, otherwise you got an exception as the failure mode?

vemv:
> If you get stuff back it works, otherwise you got an exception as the failure mode?

The side-effects of using dynapath were causing the result of `(clojure.java.classpath/classpath-directories)` to be an empty list

benedekfazekas:
will have a look

expez:
Are you married to `*warn-on-reflection*` in the `:dev` profile @bbatsov?  Personally I only turn this on when doing performance tuning (as that's the only downside of reflections).

Adding pomegranate makes the REPL quite noisy on startup now:

```
Reflection warning, cemerick/pomegranate/aether.clj:109:49 - reference to field getMessage on java.lang.Object can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:109:49 - reference to field getMessage on java.lang.Object can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:143:47 - call to method newLocalRepositoryManager on java.lang.Object can't be resolved (no such method).
Reflection warning, cemerick/pomegranate/aether.clj:173:3 - call to method setSnapshotPolicy can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:173:3 - call to method setReleasePolicy can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:181:7 - call to method addPassword on org.eclipse.aether.util.repository.AuthenticationBuilder can't be resolved (argument types: java.lang.Object).
Reflection warning, cemerick/pomegranate/aether.clj:182:7 - call to method addPrivateKey can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:179:3 - reference to field build can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:188:5 - call to method setAuthentication can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:199:34 - reference to field build can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:201:7 - call to method setProxy can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:362:5 - call to method deploy can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:381:5 - call to method install can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:470:36 - reference to field getDependency can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:476:29 - reference to field getChildren can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:614:31 - call to method setProperties can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:621:39 - reference to field getMirrorSelector can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:622:39 - call to method getMirror can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:626:7 - call to method resolveArtifacts can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:630:10 - call to method resolveVersion can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:630:10 - call to method resolveVersion can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:706:33 - reference to field getArtifact can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:706:33 - call to method setProperties can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:703:15 - call to method setArtifact can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:804:65 - reference to field getMirrorSelector can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:805:65 - call to method getMirror can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:800:31 - call to org.eclipse.aether.collection.CollectRequest ctor can't be resolved.
Reflection warning, cemerick/pomegranate/aether.clj:810:7 - call to method resolveDependencies can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:811:7 - call to method collectDependencies can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate/aether.clj:819:3 - reference to field getRoot can't be resolved.
Reflection warning, cemerick/pomegranate.clj:20:13 - call to method getDeclaredMethod can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate.clj:22:5 - call to method setAccessible can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate.clj:23:5 - call to method invoke can't be resolved (target class is unknown).
Reflection warning, cemerick/pomegranate.clj:32:17 - reference to field getParent can't be resolved.
```

bbatsov:
> Are you married to *warn-on-reflection* in the :dev profile @bbatsov? Personally I only turn this on when doing performance tuning (as that's the only downside of reflections).

I don't care about the reflection warnings, but many people do. If we disable this when we develop, then they'll suffer when we ship.

vemv:
Specifically, it can be a pain when one invokes `lein repl` in end-user projects with `warn-on-reflection` and one sees a wall of warns coming from clojure-emacs tools.

i.e. even if performance might mostly be unaffected, it's pretty noisy.

expez:
Good points! Hadn't considered that it'll look real bad when someone turns on `*warn-on-reflection*` in their own project and refactor-nrepl spews a wall of warnings.

Guess it might be prudent to fix these reflection warnings in `pomegranate` prior to cutting the 3.0.0 release.  I'm sure they'll be happy with a PR over at clj-commons fixing this.

bbatsov:
The only reason I know that people care about this is because people are constantly complaining when there are oversights in Orchard and cider-nrepl. 😆

expez:
@bbatsov I think something went wrong when you built the latest snapshot.  It's almost half the size of the 2.5.1 release.

When I inspect it all the `.class` files of the inlined deps are missing:

image

Or maybe that's the new normal with the latest mranderson release?

bbatsov:
I did a "make clean deploy", as I always do. Still, it seems the deps weren't properly inlined.

bbatsov:
Time to rebase here I guess.

expez:
Rebased.

The lint run fails and says this:

```
Possibly confusing dependencies found:
[clj-commons/pomegranate "1.2.1" :exclusions [org.clojure/clojure org.slf4j/jcl-over-slf4j org.tcrawley/dynapath]] -> [org.apache.httpcomponents/httpclient "4.5.8"]
 overrides
[clj-commons/pomegranate "1.2.1" :exclusions [org.clojure/clojure org.slf4j/jcl-over-slf4j org.tcrawley/dynapath]] -> [org.apache.maven.wagon/wagon-http "3.3.4"] -> [org.apache.maven.wagon/wagon-http-shared "3.3.4"] -> [org.apache.httpcomponents/httpclient "4.5.9" :exclusions [commons-logging]]
 and
[clj-commons/pomegranate "1.2.1" :exclusions [org.clojure/clojure org.slf4j/jcl-over-slf4j org.tcrawley/dynapath]] -> [org.apache.maven.wagon/wagon-http "3.3.4"] -> [org.apache.httpcomponents/httpclient "4.5.9" :exclusions [commons-logging]]

Consider using these exclusions:
[clj-commons/pomegranate "1.2.1" :exclusions [org.slf4j/jcl-over-slf4j org.clojure/clojure org.apache.httpcomponents/httpclient org.tcrawley/dynapath]]

Aborting due to :pedantic? :abort
Error encountered performing task 'run' with profile(s): 'base,system,user,provided,1.10,clj-kondo'
Suppressed exit
Makefile:21: recipe for target 'kondo' failed
make: *** [kondo] Error 1

Exited with code exit status 2

CircleCI received exit code 2
```

Yet locally I get no conflicting deps:

```
$ lein deps :tree
 [cider/cider-nrepl "0.25.9" :scope "provided" :exclusions [[org.clojure/clojure]]]
 [cider/orchard "0.7.1" :exclusions [[org.clojure/clojure]]]
   [org.tcrawley/dynapath "1.1.0" :exclusions [[org.clojure/clojure]]]
 [cider/piggieback "0.5.2" :scope "test" :exclusions [[org.clojure/clojure]]]
 [clj-commons/fs "1.6.307" :exclusions [[org.clojure/clojure]]]
   [org.apache.commons/commons-compress "1.20"]
   [org.tukaani/xz "1.8"]
 [clj-commons/pomegranate "1.2.1" :exclusions [[org.clojure/clojure] [org.slf4j/jcl-over-slf4j] [org.tcrawley/dynapath]]]
   [org.apache.httpcomponents/httpclient "4.5.8"]
     [commons-codec "1.11"]
     [commons-logging "1.2"]
   [org.apache.httpcomponents/httpcore "4.4.11"]
   [org.apache.maven.resolver/maven-resolver-api "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-connector-basic "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-impl "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-spi "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-transport-file "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-transport-http "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-transport-wagon "1.3.3"]
   [org.apache.maven.resolver/maven-resolver-util "1.3.3"]
   [org.apache.maven.wagon/wagon-http "3.3.4"]
     [org.apache.maven.wagon/wagon-http-shared "3.3.4"]
       [org.jsoup/jsoup "1.12.1"]
   [org.apache.maven.wagon/wagon-provider-api "3.3.4"]
   [org.apache.maven/maven-resolver-provider "3.6.1"]
     [javax.inject "1"]
     [org.apache.maven/maven-model-builder "3.6.1"]
       [org.apache.maven/maven-artifact "3.6.1"]
         [org.apache.commons/commons-lang3 "3.8.1"]
       [org.apache.maven/maven-builder-support "3.6.1"]
       [org.codehaus.plexus/plexus-component-annotations "1.7.1" :exclusions [[junit]]]
       [org.codehaus.plexus/plexus-interpolation "1.25"]
     [org.apache.maven/maven-model "3.6.1"]
     [org.apache.maven/maven-repository-metadata "3.6.1"]
     [org.codehaus.plexus/plexus-utils "3.2.0"]
     [org.slf4j/slf4j-api "1.7.25"]
 [cljfmt "0.7.0" :exclusions [[org.clojure/clojure]]]
   [com.googlecode.java-diff-utils/diffutils "1.3.0"]
   [org.clojure/tools.cli "1.0.194"]
   [rewrite-cljs "0.4.5"]
 [clojure-complete "0.2.5" :exclusions [[org.clojure/clojure]]]
 [com.google.code.findbugs/jsr305 "3.0.2" :scope "provided" :exclusions [[org.clojure/clojure]]]
 [com.google.errorprone/error_prone_annotations "2.1.3" :scope "provided" :exclusions [[org.clojure/clojure]]]
 [commons-io "2.8.0" :scope "test" :exclusions [[org.clojure/clojure]]]
 [http-kit "2.5.1" :exclusions [[org.clojure/clojure]]]
 [nrepl "0.8.3" :exclusions [[org.clojure/clojure]]]
 [org.clojure/clojure "1.9.0" :scope "provided" :exclusions [[org.clojure/clojure]]]
   [org.clojure/core.specs.alpha "0.1.24" :scope "provided"]
   [org.clojure/spec.alpha "0.1.143" :scope "provided"]
 [org.clojure/clojurescript "1.10.520" :scope "test" :exclusions [[org.clojure/clojure]]]
   [com.cognitect/transit-clj "0.8.309" :scope "test" :exclusions [[org.clojure/clojure]]]
     [com.cognitect/transit-java "0.8.332" :scope "test"]
       [com.fasterxml.jackson.core/jackson-core "2.8.7" :scope "test"]
       [org.msgpack/msgpack "0.6.12" :scope "test"]
         [com.googlecode.json-simple/json-simple "1.1.1" :scope "test" :exclusions [[junit]]]
         [org.javassist/javassist "3.18.1-GA" :scope "test"]
   [com.google.javascript/closure-compiler-unshaded "v20180805" :scope "test"]
     [args4j "2.0.26" :scope "test"]
     [com.google.code.gson/gson "2.7" :scope "test"]
     [com.google.guava/guava "25.1-jre" :scope "test"]
       [com.google.j2objc/j2objc-annotations "1.1" :scope "test"]
       [org.checkerframework/checker-qual "2.0.0" :scope "test"]
       [org.codehaus.mojo/animal-sniffer-annotations "1.14" :scope "test"]
     [com.google.javascript/closure-compiler-externs "v20180805" :scope "test"]
     [com.google.jsinterop/jsinterop-annotations "1.0.0" :scope "test"]
     [com.google.protobuf/protobuf-java "3.0.2" :scope "test"]
   [org.clojure/google-closure-library "0.0-20170809-b9c14c6b" :scope "test"]
     [org.clojure/google-closure-library-third-party "0.0-20170809-b9c14c6b" :scope "test"]
   [org.mozilla/rhino "1.7R5" :scope "test"]
 [org.clojure/data.json "2.3.1" :exclusions [[org.clojure/clojure]]]
 [org.clojure/tools.analyzer.jvm "1.1.0" :exclusions [[org.clojure/clojure]]]
   [org.clojure/core.memoize "1.0.236"]
     [org.clojure/core.cache "1.0.207"]
       [org.clojure/data.priority-map "1.0.0"]
   [org.clojure/tools.analyzer "1.0.0"]
   [org.ow2.asm/asm "5.2"]
 [org.clojure/tools.namespace "1.1.0" :exclusions [[org.clojure/clojure] [org.clojure/tools.reader]]]
   [org.clojure/java.classpath "1.0.0"]
 [org.clojure/tools.reader "1.3.5" :exclusions [[org.clojure/clojure]]]
 [pjstadig/humane-test-output "0.8.1" :exclusions [[org.clojure/clojure]]]
 [rewrite-clj "0.6.1" :exclusions [[org.clojure/clojure]]]
 [version-clj "1.0.0" :exclusions [[org.clojure/clojure]]]
```

I tried just taking the suggestion of adding `org.apache.httpcomponents/httpclient` to the list of exclusions for pomegranate, but it then fails with a `ClassNotFoundException` (which would be expected if there only was one version of this dep).

What's going on here?

vemv:
> Yet locally I get no conflicting deps:

check out how :pedantic is set. You can use `CI=true lein deps` or maybe edit temporarily project.clj

> I tried just taking the suggestion of adding org.apache.httpcomponents/httpclient to the list of exclusions for pomegranate, but it then fails with a ClassNotFoundException (which would be expected if there only was one version of this dep).

Yeah :pedantic doesn't always make an optimal suggestion. In this case, instead of an exclusion I'd add an explicit dependency on `org.apache.httpcomponents/httpclient "4.5.9"` (normally I pick the latest version when multiple are offered)

---

p.s. I self-assigned a couple issues from our tracker, I assume no overlap?

expez:
> p.s. I self-assigned a couple issues from our tracker, I assume no overlap?

Yeah, just trying to figure out why the build is failing :)

expez:
@benedekfazekas Best I can tell this is the problem:

Line 180 in `aether.clj` is this:

```clj
(defn- authentication
  [{:keys [username password passphrase private-key-file] :as settings}]
  (-> (AuthenticationBuilder.)
      (.addUsername username) <----------
      (.addPassword password)
      (.addPrivateKey private-key-file passphrase)
      .build))
```
If we check out [AuthenticationBuilder.java](
https://github.com/eclipse/aether-core/blob/master/aether-util/src/main/java/org/eclipse/aether/util/repository/AuthenticationBuilder.java#L16) it requires `javax.net.ssl.HostnameVerifier` and that isn't in the dir with renamed deps:

image

expez:
> check out how :pedantic is set. You can use CI=true lein deps or maybe edit temporarily project.clj

Thanks for the tip!  For some reason I thought `lein deps :tree` would print the same information, but only error if `pedantic?` was true...

> Yeah :pedantic doesn't always make an optimal suggestion. In this case, instead of an exclusion I'd add an explicit dependency on org.apache.httpcomponents/httpclient "4.5.9" (normally I pick the latest version when multiple are offered)

Ended up with this solution to the problem 👍

bbatsov:
What remains to be done here? Someone just asked me about the state of hotloading and I was reminded about this PR.

expez:
@bbatsov `mranderson` fails to create a viable jar.  I believe I found the problem: a `javax.*` class is renamed in an `import` clause, but not copied into the jar under the new name.  Presumably the fix would be to exclude `javax` classes from getting processed, much like we probably don't process `java.util.*`?

I tried taking a look in mranderson but it doesn't run on Windows (different path separator) and I don't have a Linux box readily available just now.

benedekfazekas:
will try to find time to have a look tmrw
--
Benedek Fazekas

benedekfazekas:
have a quick fix that I don't really like and seems that properly fixing this needs a bit more digging. meanwhile as a workaround try to exclude `javax.inject` from `clj-commons/pomegranate` deps, locally that gave me some good results, altho I had one failing test, maybe just flaky... let me know if this workaround helps

benedekfazekas:
pushed the exclusion on my own accord maybe that failing test means exclusion not such a good idea?!

benedekfazekas:
fixed the test. all seems to be well but needs a manual test

expez:
Thanks for taking a look @benedekfazekas! 

> all seems to be well but needs a manual test

I'll try to get that done this week.

@bbatsov or @benedekfazekas could you deploy a snapshot for testing? 🙏

src/refactor_nrepl/artifacts.clj
```diff
@@ -128,19 +138,88 @@
...
+        ;; A failure here isn't a big deal, it only means that resolve-missing
+        ;; isn't going to work until the namespace has been loaded manually.
+        )))
+  (slamhound/reset)
+  (slamhound-regrow/clear-cache!))
+
+(defn- parse-coordinates [coordinates-str]
+  (let [coords (try (->> coordinates-str edn/read-string (take 2) vec)
+                    (catch Exception _))]
+    (if (and (= (count coords) 2)
```
vemv, line 186: will the `2` constraint work in face of :exclusions, :classifiers etc?

src/refactor_nrepl/artifacts.clj
```diff
@@ -128,19 +138,88 @@
...
+             (symbol? (first coords))
+             (string? (second coords)))
```
vemv, line 188: Similarly, will this constraint work in face of :managed-dependencies?

With :managed-dependencies, a given member of :dependencies can be just `[foo/bar]` with no version specified.

benedekfazekas:
snapshot on clojars with version `3.0.0-pr301`

src/refactor_nrepl/artifacts.clj
expez: I don't think it makes sense to handle act on `:exclusions`.  My thinking is that you'd never need to hotload a dep that's excluded because

1. It's excluded because it's never needed
2. It gets (hot) loaded through some other artifact and was excluded to prevent ambiguity 

Finally, a request to simply load `[foo/bar]` isn't something that we should be willing to act on because it would be ambiguous.  Do they want any old version?  The latest stable?  The latest snapshot?

> :classifiers 

Didn't know about this and have never encountered it in the wild. def something we could support.

src/refactor_nrepl/artifacts.clj
expez: You'd call the op on the `:managed-dependencies` dep, and not on the `:dependencies` dep, where a version is available.

FWIW most people won't be calling hotload-deps directly on a dependency vector.  I think 99% of it's usage happens in the context of `cljr-add-project-dependency` where they've done a completed reading on artifact and then artifact version and cljr itself constructs a dependency vector to send down to refactor-nrepl

vemv: > FWIW most people won't be calling hotload-deps directly on a dependency vector. 

Wouldn't it be kind of a usual case to make a desired change over project.clj, save the file (simply so that that work isn't lost and is reused in subsequent repls) and hit a single shortcut that would load the dep based on `POINT`?

src/refactor_nrepl/artifacts.clj
vemv: Oh I think I wasn't specific enough, sorry for that

These are the sorts of inputs I had in mind:

* `[com.foo/bar "1.2.3" :exclusions [xyz]]`
* `[com.foo/bar "1.2.3" :classifier "test"]`

Both are things that can be passed straight to pomegranate IIRC. Not only we prevent a precondition from failing too easily, but we also make pomegranate resolve the right artifact, with the right exclusions (which otherwise could foil a classpath) etc

test/refactor_nrepl/artifacts_test.clj
```diff
@@ -60,3 +60,21 @@
     (is (nil? (#'artifacts/edn-read-or-nil bad-form)))
     (is (= 'foo/bar (first (#'artifacts/edn-read-or-nil good-form))))
     (is (= "1.1" (second (#'artifacts/edn-read-or-nil good-form))))))
+
+(deftest hotload-dependency-throws-exceptions
```
vemv, line 64: Gentle reminder about https://github.com/clojure-emacs/refactor-nrepl/pull/301#issuecomment-868371524 , now that we have a green build it makes sense to try it out :)

src/refactor_nrepl/artifacts.clj
expez: > make a desired change over project.clj, save the file 

`cljr-add-project-dep` does his part for you.  

I agree that we should have better support for calling `hotload-dependency` directly on a dep vector that you might have copied from elsewhere (which would then include classifiers and exclusions).

Perhaps we can just drop the validation and make sure the error message `pomegranate` throws surfaces to the user and is understandable?

vemv: Yes, SGTM :)

Might be good to have some coverage for exclusions/classifiers anyway

bbatsov:
Should we merge this one? I know there's some outstanding feedback from @vemv, but there hasn't been much movement here in the past month and it seems that the PR is an a decent shape. We can follow-up with improvements after it's merged.

vemv:
I'll add the suggested deftest, at least we should be certain we're not breaking people not using this feature (which is exactly what happened with Orchard)

Btw something might be learned from https://github.com/lambdaisland/classpath . We could keep an eye on how it evolves. It would be kind of undesirable to have N implementations of the "mutable classpath" goal in the Clojure community - it's a hard problem (particularly _across_ JDKs) so the more knowledge is reused the better.

expez:
> Should we merge this one? I know there's some outstanding feedback from @vemv, but there hasn't been much movement here in the past month and it seems that the PR is an a decent shape. 

I'm still interested in getting this fixed.  I had the misfortune of getting hit by a car during summer and so this PR has languished because I've had a cast and a concussion 🤕 Thankfully I've fully recovered and will have some time to look at this in the coming week.

> I'll add the suggested deftest, at least we should be certain we're not breaking people not using this feature (which is exactly what happened with Orchard)

Much appreciated!

vemv:
Oh man! Sorry to hear that! I had a hard 🚵‍♀️ crash 1y ago, was such a downer. Wishing you a good recovery.

Sounds good to push commits directly over this branch for a while?

bbatsov:
@expez Sorry to hear about your incident, but I'm glad you've managed to recover fully!

I asked about this PR mostly as it'd be nice to include it in the next release, but there's definitely no rush to wrap it up.

benedekfazekas:
really sorry to hear that @expez happy you recovered!

expez:
Thanks guys! 🤗 

> Sounds good to push commits directly over this branch for a while?

Yeah, feel free to push whatever you want.  

The next step I want to take is to test this end to end with clj-refactor and see if it actually does what it's supposed to.  

In the repo you linked there's a link to a [blog post](https://lambdaisland.com/blog/2021-08-25-classpath-is-a-lie) with the following quote:

> One thing I like to be able to do is add dependencies to a project without having to restart the REPL process every time. For a brief period in time this worked wonderfully for me using Pomegranate, but since Java 9 this stopped working, and my [pleas for help](https://clojureverse.org/t/dynamically-adding-to-the-classpath-in-a-post-java-9-world-help/2520) largely fell on deaf ears.

When I tested this PR in the REPL I was using ~The One True Java Version~ Java 1.8 and so I now fear that there might be some work outstanding here...

expez:
I tried the `refactor-nrepl-3.0.0-pr301` and it doesn't seem to be working unfortunately.  When the REPL comes up the middleware isn't on the classpath, but I can find any error messages anywhere.

```
[nREPL] Starting server via "c:/programdata/chocolatey/bin/lein.bat" update-in :dependencies conj ^"[nrepl/nrepl \^"0.8.3\^"]^" -- update-in :plugins conj ^"[refactor-nrepl/refactor-nrepl \^"3.0.0-pr301\^"]^" -- update-in :plugins conj ^"[cider/cider-nrepl \^"0.26.0-SNAPSHOT\^"]^" -- repl :headless :host localhost
```

image


NREPL-SERVER only has reflection warnings:

```
Warning: Clojure is excluded, assuming an appropriate fork (Clojure 1.8 or later) is provided.

Warning: Clojure is excluded, assuming an appropriate fork (Clojure 1.8 or later) is provided.

Reflection warning, nrepl/transport.clj:103:5 - call to method write can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:148:34 - call to method split can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:159:12 - call to method contains can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:173:5 - call to method startsWith can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:174:5 - call to method contains can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:175:5 - call to method contains can't be resolved (target class is unknown).
Reflection warning, nrepl/util/completion.clj:186:35 - call to method startsWith can't be resolved (target class is unknown).
Reflection warning, cider/piggieback_impl.clj:42:60 - reference to field repl_env can't be resolved.
Reflection warning, cider/piggieback_impl.clj:45:45 - reference to field repl_env can't be resolved.
Reflection warning, cider/piggieback_impl.clj:48:41 - reference to field repl_env can't be resolved.
Reflection warning, cider/piggieback_impl.clj:51:55 - reference to field repl_env can't be resolved.
Reflection warning, cider/piggieback_impl.clj:54:55 - reference to field repl_env can't be resolved.
Reflection warning, cider/piggieback_impl.clj:324:27 - reference to field repl_env can't be resolved.
Reflection warning, cider/nrepl/print_method.clj:41:1 - call to method write on java.io.Writer can't be resolved (argument types: unknown).
Reflection warning, cider/nrepl/print_method.clj:58:1 - call to method write on java.io.Writer can't be resolved (argument types: unknown).
Reflection warning, cider/nrepl/print_method.clj:86:3 - call to method write on java.io.Writer can't be resolved (argument types: unknown).
Reflection warning, C:\Users\expez\AppData\Local\Temp\form-init6309839149845217526.clj:1:9870 - reference to field deleteOnExit can't be resolved.
nREPL server started on port 59373 on host localhost - nrepl://localhost:59373

WARNING: pprint already refers to: #'clojure.core/pprint in namespace: cider.nrepl.pprint, being replaced by: #'cider.nrepl.pprint/pprint

Reflection warning, cider/nrepl/middleware/out.clj:53:22 - call to method write can't be resolved (target class is unknown).
Reflection warning, cider/nrepl/middleware/out.clj:55:24 - call to method write on java.lang.Object can't be resolved (no such method).
Reflection warning, cider/nrepl/middleware/out.clj:55:24 - call to method write on java.lang.Object can't be resolved (no such method).
Reflection warning, cider/nrepl/middleware/out.clj:57:22 - call to method write can't be resolved (target class is unknown).
Reflection warning, cider/nrepl/middleware/out.clj:59:24 - call to method write on java.lang.Object can't be resolved (no such method).
Reflection warning, cider/nrepl/middleware/out.clj:59:24 - call to method write on java.lang.Object can't be resolved (no such method).
Reflection warning, cider/nrepl/middleware/out.clj:61:21 - reference to field flush can't be resolved.
Reflection warning, cider/nrepl/middleware/out.clj:63:23 - reference to field flush on java.lang.Object can't be resolved.
Reflection warning, cider/nrepl/middleware/out.clj:63:23 - reference to field flush on java.lang.Object can't be resolved.
```

Any idea how we can get a proper error message and find out what's wrong?

vemv:
> When I tested this PR in the REPL I was using The One True Java Version Java 1.8 and so I now fear that there might be some work outstanding here...

Good note, might indicate some desirable test coverage to be added to the PR

> Any idea how we can get a proper error message and find out what's wrong?

Sorry, nothing comes to mind 😞 one thing I sometimes do in these cases is using CI as my "dev machine" i.e. I just push to it more often than usual and gather feedback from the test output. A bit slower than usual but can help.

Finally, over the last week @plexus pushed various work to cider-nrepl, Orchard etc so it seems good timing to sync up if wanting to restore the classloading feature to refactor-nrepl.

See also https://github.com/clojure-emacs/cider/issues/3049

plexus:
Thanks for looping me in @vemv. I've been getting sucked into the topic of classpath manipulation, and am increasingly convinced that we need better practices around this. If you pick any two tools/libraries from the list in https://github.com/clojure-emacs/cider/issues/3049 then I think it's more likely they won't work together than that they will.

Seems this proposal piggiebacks on Pomegranate, so it shouldn't make things worse than they are already :) I'm still trying to get a better understanding of how the different libraries/tools go about their business and how we could improve that, which I hope will result in a blog post with recommendations. Meanwhile I added some of my thoughts to https://github.com/clojure-emacs/cider/issues/3049#issuecomment-917874662

expez:
@plexus Would you implement this feature differently?  We're having a hard time making Pomegranate work as an inlined dependency, so we might as well consider other solutions.

plexus:
What are the problems with Pomegranate?

What Pomegranate does is combining two things: an Aether wrapper to resolve/download dependencies, and dynapath to add them to a `DynamicClassLoader`/`URLClassLoader`.

tools.deps add-lib is pretty similar, it has its own aether/maven wrapper, and then finds the bottom-most DynamicClassloader and appends the new jars there.

The second part is relatively straightforward, it's the resolution stuff that's complicated and where you want to rely on someone else. So you could do that with tools.deps and use dynapath or your own logic for adding the items to the classpath.

Note that tools.deps resolves transitive dependencies slightly differently than other tools/libraries, so you actually have to make a choice which resolution flavor you want. I also think for best results (consistency) you should resolve the entire set of present and additional dependencies, and only add the new entries to the classpath, instead of only resolving the one new dependency. If you do the latter you may end up with multiple versions of the same artifact on the classpath, which can cause issues.

And there are certain cases that can't be made to work properly, e.g. if someone actually wants to use this to upgrade a dependency which was present at load time. You can add the new version, but the old version will remain on the classpath (it can't be removed), **and** it will actually get precedence when doing resource lookups (as I've described [in the blog post about classloaders](https://lambdaisland.com/blog/2021-08-25-classpath-is-a-lie). You might want to look at this and at how [lambdaisland.classpath](https://github.com/lambdaisland/classpath/blob/cb786e9956ea169dd1ef50fc5c29959e9025cff8/src/lambdaisland/classpath.clj#L342) does it for some inspiration.

expez:
> What are the problems with Pomegranate?

None, really.  It just interacts badly with the source rewriting tool we use to prevent conflicts on the classpath.

Thanks for confirming that we are on the right track!

bbatsov:
I think it’s fine to not in-line it. After all, how many projects have dependencies like pomegranate? That would make sideloading refactor-nrepl a bit harder, though.

benedekfazekas:
we don't really have a mranderson problem atm, do we?

vemv:
(There's https://github.com/jafingerhut/dolly as well if anything else fails 😄, it's doing a decent job for Eastwood)

expez:
> we don't really have a mranderson problem atm, do we?

When I start a REPL using the jar you published the REPL comes up without refactor-nrepl on the classpath.  I can't see any errors anywhere so I've no idea what's wrong.  The mrandersoned tests also run fine on the CI server.  Pretty frustrating failure mode.

benedekfazekas:
> When I start a REPL using the jar you published the REPL comes up without refactor-nrepl on the classpath. I can't see any errors anywhere so I've no idea what's wrong. The mrandersoned tests also run fine on the CI server. Pretty frustrating failure mode.

oh i see your comment above now. would be interesting to see the line how your REPL is started up, what is injected etc. there could be a problem there not with the actual jar

expez:
@benedekfazekas posted that too above, but here it is again:

```
[nREPL] Starting server via "c:/programdata/chocolatey/bin/lein.bat" update-in :dependencies conj ^"[nrepl/nrepl \^"0.8.3\^"]^" -- update-in :plugins conj ^"[refactor-nrepl/refactor-nrepl \^"3.0.0-pr301\^"]^" -- update-in :plugins conj ^"[cider/cider-nrepl \^"0.26.0-SNAPSHOT\^"]^" -- repl :headless :host localhost
```

benedekfazekas:
> @benedekfazekas posted that too above, but here it is again:
> 
> ```
> [nREPL] Starting server via "c:/programdata/chocolatey/bin/lein.bat" update-in :dependencies conj ^"[nrepl/nrepl \^"0.8.3\^"]^" -- update-in :plugins conj ^"[refactor-nrepl/refactor-nrepl \^"3.0.0-pr301\^"]^" -- update-in :plugins conj ^"[cider/cider-nrepl \^"0.26.0-SNAPSHOT\^"]^" -- repl :headless :host localhost
> ```

you did, sorry. and it looks fine too. will check whats in that jar

benedekfazekas:
only had time for a quick check now but doing this on a random project
```

**Comment by benedekfazekas** [comment]
I don't think there is a problem with `3.0.0-pr301` jar @expez 

I jacked it with the following params:
```
;;  Startup: /usr/local/bin/clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.9.0-beta2"} refactor-nrepl/refactor-nrepl {:mvn/version "3.0.0-pr301"} cider/cider-nrepl {:mvn/version "0.26.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[refactor-nrepl.middleware/wrap-refactor,cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: cider.nrepl.inlined-deps.orchard.v0v7v1.orchard.misc, being replaced by: #'cider.nrepl.inlined-deps.orchard.v0v7v1.orchard.misc/update-vals
WARNING: update-keys already refers to: #'clojure.core/update-keys in namespace: cider.nrepl.inlined-deps.orchard.v0v7v1.orchard.misc, being replaced by: #'cider.nrepl.inlined-deps.orchard.v0v7v1.orchard.misc/update-keys
Their versions are 2.5.1 (package: 20210628.1154) and 3.0.0-pr301, respectively.
You can mute this warning by changing cljr-suppress-middleware-warnings.WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-vals
WARNING: update-keys already refers to: #'clojure.core/update-keys in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-keys
WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-vals
WARNING: update-keys already refers to: #'clojure.core/update-keys in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-keys
WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.passes, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-vals
WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.passes.uniquify, being replaced by: #'refactor-nrepl.inlined-deps.toolsanalyzer.v1v0v0.clojure.tools.analyzer.utils/update-vals
```

and successfully run find-usages on my local machine. let me know if I missed anything here...

**Comment by expez** [comment]
@benedekfazekas 

On ubuntu 20.04, with openjdk 16.

```
$ clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.9.0-beta2"} refactor-nrepl/refactor-nrepl {:mvn/version "3.0.0-pr301"} cider/cider-nrepl {:mvn/version "0.26.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[refactor-nrepl.middleware/wrap-refactor,cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
Syntax error (FileNotFoundException) compiling at (refactor_nrepl/core.clj:1:1).
Could not locate clojure/tools/namespace/parse__init.class, clojure/tools/namespace/parse.clj or clojure/tools/namespace/parse.cljc on classpath.
```

 I tested the equivalent command on win 10 with jdk8 and jdk 17 and got the same result.

**Comment by benedekfazekas** [comment]
this is what i see in the ns decl of `refactor-nrepl.core`:

```

expez:
> my only guess is that you tried to build the jar locally and it went wrong, or that there was a previous version of the same jar that was broken. maybe try to redownload the jar?

I hadn't built the jar locally and I've downloaded the jar on two different machines.  I deleted it manually just to be sure one last time, with the same result.

I also tried building this branch locally, without inlining Pomegranate, and after installing the jar locally (under a different name :stuck_out_tongue:) I get a repl without any refactor-nrepl deps on the classpath again.

I must be doing something wrong, but I'm not sure what that is at this point.  Maybe someone else can take this across the finish line...

benedekfazekas:
it is equally possible that i am doing something wrong but can't figure out what that would be... 😖 we are bit stuck here, i agree. @vemv @bbatsov any chance you can have a look if this jar **does** or does **not** work for you...

bbatsov:
It works for me with JDK 14. I'll install and test 16 later.

bbatsov:
Same with JDK 16. I ran:

```
$ clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.9.0-beta2"} refactor-nrepl/refactor-nrepl {:mvn/version "3.0.0-pr301"} cider/cider-nrepl {:mvn/version "0.26.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[refactor-nrepl.middleware/wrap-refactor,cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
```

And then did a `cider-connect`. Everything worked fine.

expez:
Thanks for testing @bbatsov.

I still can't get a REPL going in emacs, but connecting to this:

```
$ clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "0.9.0-beta2"} refactor-nrepl/refactor-nrepl {:mvn/version "3.0.0-pr301"} cider/cider-nrepl {:mvn/version "0.26.0"}} :aliases {:cider/nrepl {:main-opts ["-m" "nrepl.cmdline" "--middleware" "[refactor-nrepl.middleware/wrap-refactor,cider.nrepl/cider-middleware]"]}}}' -M:cider/nrepl
```

does work, if I don't do it in the context of the `refactor-nrepl` project.  

* `pomegate` works as it should on Java 16 (tested at the REPL).
* hotload-deps in the `3.0.0-pr301` artifact fails with `The super POM /refactornrepl300pr301/org/apache/maven/model/pom-4.0.0.xml was not found, please verify the integrity of your Maven installation` during dependency resolution
* I tried building a jar from this branch without `pomegranate` as an inlined dep (using `lein do clean, inline-deps && lein with-profile +plugin.mranderson/config install`) but that gives me a jar without pomegranate in it which obv fails when `hotload-deps` is called

Might spend some more time on this tomorrow, but kind of stuck again.  The dependency resolution fails deep int the java code of the maven-resolver AFAICT.

expez:
I was at a bit of a loss as to how I could make `pomegranate` work so I decided to give `add-libs` a try instead.

I got it working at the REPL, but after packaging it up it fails because `mranderson` doesn't handle these [loads](https://github.com/clojure/tools.deps.alpha/blob/master/src/main/clojure/clojure/tools/deps/alpha.clj#L815-L820)
```clj
;; Load extensions
(load "/clojure/tools/deps/alpha/extensions/maven")
(load "/clojure/tools/deps/alpha/extensions/local")
(load "/clojure/tools/deps/alpha/extensions/git")
(load "/clojure/tools/deps/alpha/extensions/deps")
(load "/clojure/tools/deps/alpha/extensions/pom")
```

Because e.g. the maven extension is now called `refactor-nrepl.inlined-deps.toolsdepsalpha.v0v12v1048.clojure.tools.deps.alpha.extensions.maven`

Since the `add-libs` code is unreleased and lives on its branch I opted to just include it in our jar.  Initially I wanted to preserve the namespace but when I stored it under `src/clojure/tools/deps/alpha/repl.clj` `mranderson` didn't pick it up, so I moved it where it is now :shrug: 

Oh, and I also had to [edit one line](https://github.com/clojure-emacs/refactor-nrepl/pull/301/commits/3ea4fc388c15699e67cc098990a1577ee18316b6#diff-89db1fe128af16b03f34bc78842c409d84e02476a9bcff429c837d9aeb144e0eR83) in `add-libs` to make it work with projects that are leiningen based.  This probably means that projects that aren't `tools.deps` based will be less robust to weird edge-cases, but given the intended use-case I think that's OK.

It'll probably be a week or two until I have time to look at this again :disappointed: 

If anyone wants to pick up the torch, feel free to push directly to this branch,

expez:
> Are there any updates on this topic?

Nope.  I've had no time to work on this.  IIRC I ran into some problems with `mranderson` that I was unable to solve. `tools.deps` pulls in a *huge* amount of java deps, so it's not entirely surprising that we ran into some edge-cases when doing code rewriting.

I believe this feature is sorely missed, so if you want to get involved that would be most welcome!

plexus:
For what it's worth we've taken a new take on hot reloading of dependencies with [lambdaisland/classpath](https://github.com/lambdaisland/classpath), and expanded on it in [launchpad](https://github.com/lambdaisland/launchpad). This will watch `deps.edn` for changes, as well as the `deps.edn` of any projects linked with `:local/root`. It works by leveraging tools.deps, and injecting a custom classloader which can "override" entries from the classpath the system started with.

Not saying it's flawless, as mentioned there are a lot of tricky edge cases because a lot of JVM stuff doesn't expect you to mess with the classpath mid-flight. E.g. adding ClojureScript libs still requires a restart. But it's been working pretty well for us.

I also did a blog posts about this stuff. It's hairy: [The Classpath is a lie](https://lambdaisland.com/blog/2021-08-25-classpath-is-a-lie).

PR #303, opened 2021-06-27
Introduce `:insert-newline-after-require`

vemv:
* Upgrade Eastwood, clj-kondo
* Makefile: also use `-user` profile
  * This increases reproducibility while developing the project.
* Introduce `:insert-newline-after-require` option
  * Context: https://github.com/clojure-emacs/refactor-nrepl/pull/300
  
### DIFF
--- a/test/resources/artifacts_pprinted
+++ b/test/resources/artifacts_pprinted
@@ -11,4 +11,5 @@
     [transport :as transport]]
    [org.httpkit.client :as http]
    [refactor-nrepl.externs :refer [add-dependencies]])
-  (:import java.util.Date))
+  (:import
+   java.util.Date))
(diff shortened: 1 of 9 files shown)


test/resources/artifacts_pprinted
```diff
@@ -11,4 +11,5 @@
     [transport :as transport]]
    [org.httpkit.client :as http]
    [refactor-nrepl.externs :refer [add-dependencies]])
-  (:import java.util.Date))
+  (:import
```
vemv, line 14: I made the treatment of :import homogeneous relative to that of :require, as most likely users never want an asymmetry.

(It's also what https://github.com/bbatsov/clojure-style-guide/tree/e6cb6dab0925bcfabcd4ec96a2f69f41bf9c698e#line-break-ns-declaration suggests in its example)

bbatsov:
Great work! Thanks!

expez:
Great work @vemv.

You might already be on this, but we also need a `defvar` and a changelog entry in clj-refactor.el to toggle this on/off and then send it to the middleware.  We already have a setup for this so should be fairly easy.

vemv:
> You might already be on this, but we also need a defvar and a changelog entry in clj-refactor.el to toggle this on/off and then send it to the middleware. 

Yeah I figured so! Later today

PR #307, opened 2021-06-27
[Fix #305] No newline after :as or :refer

### LINKED ISSUE #305: Prevent clean-ns from putting :as and :refer on their own lines
This is due the default output from `clojure.pprint/pprint` when a vector grows long.  It will then put each element on their own line.  This looks especially stupid with libspec vectors:

```clj
[a.pretty.long.ns.gets.its.own.line.like.it.should
 :as 
 but-then-the-alias-comes-way-down-here]
```

A better formatting would be:

```clj
[a.pretty.long.ns.gets.its.own.line.like.it.should
 :as the-alias-comes-on-this-line-instead]
```

I want to give `:refer` the same treatment.  

```clj
[a.pretty.long.ns.gets.its.own.line.like.it.should
 :refer
 [this that]
```

becomes:

```clj
[a.pretty.long.ns.gets.its.own.line.like.it.should
 :refer [this that]
```

This will, however, cause a bit of a regression if you have a ton of referred symbols:

```clj
(require
 '[a.pretty.long.ns.gets.its.own.line.like.it.should
   :refer
   [this
    that
    the
    other
    more
    moar
    and
    finally
    the
    problem
    on
    comes
    on
    line-wrap]])
```

becomes:

```clj
(require '[a.pretty.long.ns.gets.its.own.line.like.it.should
           :refer [this
                   that
                   the
                   other
                   more
                   moar
                   and
                   finally
                   the
                   problem
                   on
                   comes
                   on
                   line-wrap]])
```

I think these both look super bad.  This behavior will privilege short `:refer` clauses, but I think that's fine.  Having tons of referred symbols is an anti-pattern anyway, so we should nudge people toward not doing that. 

If someone does this consistently we can always take another PR improving this edge-case further.

expez:
This makes the ns form take up less vertical space and makes it easier
on the eyes.

This was never done intentionally, but a side-effect of simply delegating
the printing of regular libspecs to `clojure.core.pprint`, which puts
each element of a vector on its own line when the line gets long enough.

### DIFF
--- a/src/refactor_nrepl/ns/pprint.clj
+++ b/src/refactor_nrepl/ns/pprint.clj
@@ -14,7 +14,7 @@
   (vec (concat (remove sequential? libspecs)
                (filter sequential? libspecs))))
 
-(defn- pprint-prefix-form [[name & libspecs]]
+(defn- pprint-libspec-with-prefix-form [[name & libspecs]]
   (printf "[%s" name)
   (let [ordered-libspecs (libspec-vectors-last libspecs)]
     (dorun
@@ -35,10 +35,21 @@
                             (printf "%s " libspec))))))
                   ordered-libspecs))))
 
-(defn insert-clause-delimiter []
+(defn- insert-clause-delimiter []
   (if (:insert-newline-after-require *config*)
     (println)
     (print " ")))
+
+(def ^:const ^:private as-or-refer-re-pattern
+  (re-pattern (str "(:as|:refer)" (System/lineSeparator))))
+
+(defn- pprint-libspec [libspec]
+  ;; If a vector gets too long `pprint` will print one element per line.
+  ;; This puts `:as` and `:refer` on their own line, which causes the ns form
+  ;; to take up too much vertical space.
+  (printf (str/replace (with-out-str (pprint libspec))
+                       as-or-refer-re-pattern
+                       "$1")))
 
 (defn pprint-require-form
   [[_ & libspecs]]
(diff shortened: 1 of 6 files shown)


src/refactor_nrepl/ns/pprint.clj
```diff
@@ -35,11 +35,19 @@
...
   (if (:insert-newline-after-require *config*)
     (println)
     (print " ")))
 
+(defn- pprint-libspec [libspec]
+  ;; If a vector gets too long `pprint` will print one element per line.
+  ;; This puts `:as` and `:refer` on their own line, which causes the ns form
+  ;; to take up too much vertical space.
+  (printf (str/replace (with-out-str (pprint libspec))
+                       (re-pattern (str "(:as|refer)" (System/lineSeparator)))
```
vemv, line 48: `(re-pattern (str "(:as|refer)" (System/lineSeparator))` will be invoked once per `pprint-libspec` call, can be optimized

vemv, line 48: Looks like `refer` (no colon) would be matched, might fail in some corner case

expez: Fixed, thanks!

expez: It's less efficient for sure, but the code is much easier to work with this way.

Since `n` is very low (nobody has hundreds of libspecs) the performance difference will be very small.  I'd rather leave this as is and prioritize readability/maintainability.

vemv: Tbh this doesn't like a huge compromise wrt maintainability:

```clj
(def clause-re (re-pattern (str "(:as|refer)" (System/lineSeparator))
```

> Since n is very low (nobody has hundreds of libspecs)

I don't know, think of doing a project-wide refactor over 1000 files, each having 10 libspecs as a conservative average.

Saving 1s or such can be nice.

vemv (approved):
LGTM!

src/refactor_nrepl/ns/pprint.clj
expez: Think I misunderstood a little; thanks for elaborating.  You're right of course 🙂 

Fixed.

PR #308, opened 2021-06-27
Change default to no prefix re-writing

expez:
The default has already been changed in clj-refactor so this is
essentially a no-op, but makes the change more explicit for those
working on the middleware.

### DIFF
--- a/src/refactor_nrepl/config.clj
+++ b/src/refactor_nrepl/config.clj
@@ -13,7 +13,7 @@
    :prune-ns-form true
 
    ;; Should `clean-ns` favor prefix forms in the ns macro?
-   :prefix-rewriting true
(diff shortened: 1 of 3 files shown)


bbatsov:
There's a linting failure.

expez:
> There's a linting failure.

Fixed, hopefully.  My emacs settings are really working against me with this linter :cry

vemv (approved):
lgtm!

bbatsov:
@expez You can simply update cljfmt's settings if something's off. That's what I do for the other projects - e.g. https://github.com/clojure-emacs/cider-nrepl/blob/master/project.clj#L133

bbatsov:
I'll push a new snapshot now.

PR #309, opened 2021-06-27
 [Fix #306] Replace cheshire with data.json 

### LINKED ISSUE #306: Drop cheshire for clojure.data.json
We've had some issues with `cheshire` in the past.  It's also a library with a ton of dependencies compared to `clojure.data.json` which is entirely self-contained.  

`clojure.data.json` recently went through a round of performance tuning and is now faster than `cheshire` on some payloads and within spitting distance on most others.

The big payload from clojars is read as `edn` and `cheshire` is only used on the comparatively small payloads for maven central. I expect this change to:

1. Making parse an insignificant amount faster
2. Make our jar smaller 
3. Make it less likely that we cause dependency problems in the future

expez:


expez:
Any idea why half the build matrix failed?  Guess I'll try to rebuild and hope it goes away...

bbatsov:
Nothing comes to mind immediately - I see that according to the errors piggieback's not loading, but  I wonder why would something like this happen.

expez:
There's this comment in `.circleci/config.yaml`:

```
              # TODO: add "master" (tests would fail due to piggieback)
```

It's referring to the `master` branch of clojure, I think.  Does that bring back any memories?  Guess it's unlikely to be related 🤷

vemv:
> [...] It's referring to the master branch of clojure, I think. 

Yes, it does https://github.com/clojure-emacs/refactor-nrepl/commit/1dfb5c125a8411a2c0919539a30e11c014603ae3

In that commit I simply changed a commented-out Circle step for that one comment. I did observe that building "master" would fail for due to piggieback.

I don't know at all why, though

vemv:
There isn't any discernible pattern as for why certain jobs fails and others don't. That invites me to believe that there's some kind of mixed-cache issue.

This might be a fix?

```diff
      - with_cache:
-          cache_version: << parameters.clojure_version >>
+          cache_version: << parameters.clojure_version >> << parameters.jdk_version >>
```

vemv:
No luck :/

I'm trying another thing though, looks promising

vemv:
@expez: I will create this morning a separate PR proposing various things that would fix the piggieback issue

expez:
Closing this in favor of #315

PR #310, opened 2021-06-27
[Fix #239] Teach clean-ns about more metadata 

### LINKED ISSUE #239: clean-ns: false positives for symbol usage

expez:
The performance hit in considering literally every piece of metadata
shouldn't be too detrimental so we'll just err on the side of caution
for now.

### DIFF
--- a/src/refactor_nrepl/find/symbols_in_file.clj
+++ b/src/refactor_nrepl/find/symbols_in_file.clj
@@ -34,6 +34,12 @@
               (.substring s 0 (dec (.length s)))
               s))))
 
+(defn- strip-meta-from-reader
+  [metadata]
+  (-> metadata
+      (dissoc :line :column :end-line :end-column)
+      not-empty))
+
 (defn symbols-in-file
   "Return a set of all the symbols occurring in the file at path.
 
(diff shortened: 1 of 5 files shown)


test/resources/ns_with_lots_of_meta.clj
```diff
@@ -1,12 +1,18 @@
...
+   [clojure.string :as s]
+   [clojure.pprint :refer [fresh-line]]))
 
 (defn useit
   []
   (fresh-line))
+
+(defn issue-239-metadata-based-protocol-extension []
+  ;; https://github.com/clojure-emacs/refactor-nrepl/issues/239
+  ^{`s/whatever-goes true} {})
```
vemv, line 18: Could be nice to also exercise a case with metametadata:

```clj
(defn foo ^{:foo ^:bar []} [])
```

I think this PR already handles that as I see some recursion, so maybe no impl changes will be even needed

vemv (commented):
Thanks for this one!

src/refactor_nrepl/find/symbols_in_file.clj
```diff
@@ -63,24 +69,33 @@
...
-                set)))))))
+               syms (atom #{})]
+           (letfn [(collect-symbols [form]
+                     (let [minimal-meta (strip-meta-from-reader (meta form))]
+                       ;; metadata used in metadata-based protocol extensions
+                       ;; see #239 for an example
+                       (when minimal-meta
+                         (->> minimal-meta
+                              seq
+                              flatten
```
vemv, line 80: `flatten` is somewhat fragile:

```
> (flatten #{1 2 3})
()
```

You call seq right before so my example can't happen, but still :)

So, I think you can simply call `(prewalk collect-symbols minimal-meta)` here

test/resources/ns_with_lots_of_meta.clj
expez: Yeah, it will handle that, but the stack will blow if you go all Inception on your source code with meta levels 🙄

src/refactor_nrepl/find/symbols_in_file.clj
expez: This should be fine as it's just flattening a map via `seq`.

Didn't know `flatten` behaved like this.  Thanks for the heads up!

.circleci/config.yml
```diff
@@ -157,10 +157,6 @@ workflows:
       - util_job:
           name: Code Linting
           steps:
-            - run:
```
bbatsov, line 160: I think dropping the linter is a mistake, as I know from experience that many people use Emacs, but don't run the necessary formatting commands in it. I constantly get PRs with misaligned code and `cljfmt` has been instrumental in warding those off. Not a big deal, but I'd rather stay consistent across all of our projects in terms of lint checks.

expez: Aight, I dropped the commit.

Was going to suggest we turn it off temporarily, as I use aggressive indent mode and ~~fuck up~~ improve the indentation constantly.  Then it occurred to me that it would be some work to turn it back on again, after the holidays when I'll have little time to work on refactor-nrepl 😞

bbatsov:
Seems you'll need to rebase here.

expez:
Done

test/resources/ns_with_lots_of_meta.clj
```diff
@@ -1,12 +1,18 @@
-(ns ^:md1 ^:md2 ^{:longhand "as well" :really? "yes" :ok true} ns-with-gen-class-methods-meta
+(ns ^:md1 ^:md2 ^{:longhand "as well" :really? "yes" :ok true}
+ ns-with-gen-class-methods-meta
   (:gen-class
    :methods [^:static [foo [String] String]
              ^{:test true} [bar [String] String]
              ^{:other "text"} [baz [String] String]]
    :name Name)
-  (:require [clojure.string :as s]
```
vemv, line 7: I see `clojure.string` was already present here as of `master`.

This means this PR changes the intent of `[clojure.string]` from "should be cleaned up" to "should not be cleaned up".

This results in a loss of test coverage; I'd suggest adding an extra libspec so that both things will be exercised

expez: Yeah, but that's fine.  We have tons of other tests that tests removal.  I added another ns form that gets removed 🤷

vemv (approved):

PR #311, opened 2021-06-28
[#Fix 289] Namespaced keywords vs find (local) symbol 

expez:
The edge-case handling for local symbols in opt-maps wasn't using a
reader that was configured correctly and choked on a namespaced keyword.

This PR will also drop support for clojure 1.8:

Destructuring with namespaced keywords was added in clojure 1.9 and
this feature is now referenced in our tests.

According to the state of Clojure 2021 survey only 6.36% of users are
still using this version for development or production.  Since Clojure
is a provided dep in this project they can continue to due so but
'official support' ends as we move to version 3.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Bugs fixed
+* [#289](https://github.com/clojure-emacs/refactor-nrepl/issues/289): Fix an edge-case with involving keywords that caused find-symbol to crash.
 * [#305](https://github.com/clojure-emacs/refactor-nrepl/issues/305): Don't put `:as` or `:refer` on their own lines in the ns form, when the libspec is so long it causes the line to wrap.
 * [clojure-emacs/clj-refactor.el#459](https://github.com/clojure-emacs/clj-refactor.el/issues/459): `clean-ns` should conform to the style guide: `(:require` in the ns form should be followed by a newline.
   * You can opt out via the new `:insert-newline-after-require` configuration option.
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -25,7 +25,11 @@
 (defn ns-from-string
   "Retrieve the symbol naming the ns from file-content."
   [file-content]
-  (second (parse/read-ns-decl (PushbackReader. (java.io.StringReader. file-content)))))
+  (-> file-content
+      java.io.StringReader.
+      PushbackReader.
+      parse/read-ns-decl
+      second))
 
 (defn ns-name-from-readable
   "Call slurp on readable and extract the ns-name from the content."
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -380,3 +380,12 @@
   [symbol-or-keyword]
   (when (prefix symbol-or-keyword)
     symbol-or-keyword))
+
+(defmacro with-clojure-version->=
+  "Guard the evaluation of `body` with a test on the current clojure version."
+  {:style/indent 1}
+  [{:keys [major minor] :as _clojure-version} & body]
+  (when (or (> (:major *clojure-version*) major)
+            (and (= (:major *clojure-version*) major)
+                 (>= (:minor *clojure-version*) minor)))
+    `(do ~@body)))
--- a/test/refactor_nrepl/integration_tests.clj
+++ b/test/refactor_nrepl/integration_tests.clj
@@ -3,6 +3,7 @@
             [clojure.test :refer :all]
             [nrepl.server :as nrepl]
             [refactor-nrepl middleware
+             [analyzer :as analyzer]
              [client :refer :all]
              [core :as core]]
             [clojure.string :as str])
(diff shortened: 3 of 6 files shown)


expez:
@bbatsov Where is the list of required checks to make the build pass?  I thought I got everything out of `.circleci/config/yaml` but maybe I missed something?

vemv:
It's set up in https://github.com/clojure-emacs/refactor-nrepl/settings/branch_protection_rules/2122093, I updated it just now (which could be reverted if we found out 1.8 is worth supporting... don't think so!)

test/refactor_nrepl/integration_tests.clj
```diff
@@ -206,6 +207,14 @@
           result (remove keyword? response)]
       (is (= 3 (count result)) (format "expected 3 results but got %d" (count result))))))
 
+(deftest find-local-in-namespaced-destructuring
```
vemv, line 210: We could do quite simply:

```clj
(when (modern? *clojure-version*)
  (deftest ...)
```

It seems worth it since other than this deftest there seemingly isn't another strong reason to drop 1.8 support.

I think keeping 1.8 support is a good idea. Many modern projects still use 1.8 or such by default, and 1.10 will be defined in a separate profile. For those cases, refactor-nrepl will run in an 1.8 environment by default.

expez: Might as well drop it now that we are bumping our major version.  

As I mentioned in the commit message people can still run 1.8 if they want to, as the clojure version is managed by the user.

vemv: There's the nuance that dropping 1.8 from CI would mean we would no longer detect incompatibilities in advance.

bbatsov:
Dropping 1.8 will also mean that cider-nrepl and refactor-nrepl will be misaligned in terms of supported Clojure versions. I don't plan to drop support for 1.8 in cider-nrepl until Clojure 1.11 gets released and widely adopted.

bbatsov:
(and yeah - I get you mean this only for the CI, but I agree with @vemv that this would mean we won't notice any regressions until something's in the wild)

expez:
I dropped the commit removing clojure 1.8 support, although I disagree strongly with expending limited volunteer time on something that provides no value to the world.

Unfortunately the guard around the code doesn't seem to work.  Is it obvious to either of you why it doesn't?

test/resources/testproject/src/com/example/five.clj
```diff
@@ -58,6 +58,15 @@
     [:bar :foo]
     (count foo)))
 
+(when-let [{:keys [minor major]} *clojure-version*]
+  (and (>= major 1)
```
vemv, line 62: this `and` is outside a condition

(if fixing this still doesn't work, then one would have to create a new `resources` dir just for this case. It would be added to :resource-paths only when using the :1.9 and :1.10 profiles)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Bugs fixed
+* [#289](https://github.com/clojure-emacs/refactor-nrepl/issues/290): Fix an edge-case with involving keywords that caused find-symbol to crash.
```
vemv, line 6: it points to 290... off by one :)

expez: Thanks, fixed.

test/resources/testproject/src/com/example/five.clj
expez: Thanks, still working on that first cup of coffee :)

expez:
Btw I've bumped the provided clojure dep in the dev profile.  The reason for this is that you can't run all the tests during development otherwise.  If a mistake is made, and a feature from a more recent release is used, the CI server will catch it.

vemv:
Looks like the build keeps failing for the same reason, I'd give a shot to the alt given in https://github.com/clojure-emacs/refactor-nrepl/pull/311#discussion_r660383510

expez:
> Looks like the build keeps failing for the same reason

Yeah, makes sense when I think about it.  We're guarding against evaluation but the reader still gets the opportunity to choke on syntax it doesn't understand.

>  I'd give a shot to the alt given in #311 (comment)

I added a macro to prevent the reader from seeing the code at all.  Think that's a bit easier to work with than the separate resource paths.

Sorry about the noise, if you're tracking build failures @vemv but I can't run all the tests locally... There's [this issue](https://github.com/clojure-emacs/refactor-nrepl/issues/312) and when I try to do <kbd> M-x cider-test-run-ns-tests </kbd> i just get `No test namespace: refactor-nrepl.integration-tests-test`.  It's much easier to just push to the CI server than manually run each and every test in the project 🙄

src/refactor_nrepl/core.clj
```diff
@@ -376,3 +380,11 @@
   [symbol-or-keyword]
   (when (prefix symbol-or-keyword)
     symbol-or-keyword))
+
+(defmacro with-clojure-version->=
+  "Guard the evaluation of `body` with a test on the current clojure version."
+  {:style/indent 1}
+  [{:keys [major minor] :as _clojure-version} & body]
+  (when (and (>= (:major *clojure-version*) major)
+             (>= (:minor *clojure-version*) minor))
```
vemv, line 389: I think that if `*clojure-version*` is 2.0, then the minor comparison can make the condition fail

IOW if major is > (but not >=), then minor is irrelevant?

expez: Fixed this, although I doubt we'll ever see a clojure 2.0 (would break their commitment to backward compatibility)

vemv (approved):
LGTM!

PR #313, opened 2021-06-29
[Fix #312] Local tests without mranderson

### LINKED ISSUE #312: Make the tests run locally again under Windows
This snippet:

```clj
(ns global-test-setup
  "This namespace's contents will be automatically loaded by `lein test` (or any test runner)."
  (:require
   [refactor-nrepl.inlined-deps.toolsnamespace.v1v1v0.clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
```

references an ns that doesn't exist until mranderson has run.

My current side project is Windows only and mranderson doesn't run on Windows.  The presence of his ns therefore
makes it impossible to run the tests locally.  Using the CI server to test that everything is OK makes the feedback loop quite a bit slower.

Why is this ns referencing a namespace that doesn't exist?

I also don't get this part:

```clj
(when (System/getenv "CI") ; don't alter local environments (which may also use clojure.tools.namespace.repl)
  ;; Intentionally exclude the `lein-plugin/` source path:
  (set-refresh-dirs "src" "test"))
```

Why would we need a refresh dir to be set on the CI server instead of locally?  It looks like this NS is a noop as written.  Can we just remove this file?

expez:


expez:
Does this look like it'll have the same effect, @vemv?

test/global_test_setup.clj
```diff
@@ -1,7 +1,12 @@
 (ns global-test-setup
   "This namespace's contents will be automatically loaded by `lein test` (or any test runner)."
   (:require
-   [refactor-nrepl.inlined-deps.toolsnamespace.v1v1v0.clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
+   [clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
+
+(when (find-ns
```
vemv, line 6: find-ns can return nil if nothing required the ns before. i.e. it's not a reliable way to detect whether a ns exists

I'd recommend to `(try (require ...))` directly instead

vemv (commented):
Approach LGTM, with just one nuance

test/global_test_setup.clj
expez: Good idea, thanks! 👍

test/global_test_setup.clj
```diff
@@ -1,7 +1,11 @@
 (ns global-test-setup
   "This namespace's contents will be automatically loaded by `lein test` (or any test runner)."
   (:require
-   [refactor-nrepl.inlined-deps.toolsnamespace.v1v1v0.clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
+   [clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
```
vemv, line 4: Maybe move this addition to the `catch`? Else an ambiguity is created

expez: 👍

vemv (approved):

expez:
Thanks a lot for your help with this @vemv!

PR #314, opened 2021-06-30
Fix a test

vemv:
It used a single-handed `=`: `(= (some #(= % 'refactor-nrepl.ns.namespace-aliases) ...`, which was making the test invalid.

While I was there, I also DRYed related deftests.


expez (approved):
Nice 👍

PR #315, opened 2021-07-01
Strengthen the project/codebase

vemv:
As the https://github.com/clojure-emacs/refactor-nrepl/pull/309 thread shows, the project setup / codebase can be occasionally hard to debug.

This PR strengthens various things.

The first commit is particularly important: it splits `test/` into `test`, `test-resources`, `java-test`. This resembles more a traditional structure (vs. a nested one) that various tools can work better with.

There were many signs IMO that the prior layout was wrong e.g. words like `resources` were part of the sample ns names.

Other things are also fixed; I found out about those while trying to get the build green. The commit history in this PR is granular (w/ extended messages etc) so I won't repeat what's been done here :)

These are some nifty things that were enabled as a result:

* Ability to run Eastwood in `test`
  * (it found https://github.com/clojure-emacs/refactor-nrepl/pull/314 yesterday)
* Exercise clojure @ master again
* Replace cheshire with data.json with a green build
  * I included @expez 's commit here.

I also believe end-users will have a reduced chance for issues, with e.g. the introduced extra cautions around `find-ns`.


src/refactor_nrepl/util.clj
```diff
@@ -52,3 +52,17 @@
...
+  (or (-> candidate str (.startsWith "deps."))
+      (-> candidate str (.startsWith "mranderson"))
+      (-> candidate str (.startsWith "eastwood.copieddeps"))))
+
+(defn- refactor-nrepl? [candidate]
+  ;; the refactor-nrepl implmentation should never analyse itself w/ tools.analyzer, as this can cause opaque issues:
+  (and (-> candidate str (.startsWith "refactor-nrepl"))
+       (not (-> candidate str (.contains "test")))))
+
+(def invalid-fqn?
```
expez, line 66: I don't like using `def` to create functions because I then lose the editor typehint and I use those a lot to remind me what sort of arg(s) are expected.

Maybe we can give this a name like `self-referential?` or something to that effect?  My thinking is that this isn't a generally invalid fqn, but invalid in the context of refactor nrepl.

src/refactor_nrepl/find/find_symbol.clj
```diff
@@ -136,7 +137,12 @@
                                (str/join "/" [namespace var-name]))
         referred-syms (libspecs/referred-syms-by-file&fullname)]
     (->> (core/dirs-on-classpath)
-         (mapcat (partial core/find-in-dir (some-fn core/clj-file? core/cljc-file?)))
+         (mapcat (partial core/find-in-dir (every-pred (some-fn core/clj-file? core/cljc-file?)
+                                                       (fn [f]
```
expez, line 142: `core/self-referential-file?`

src/refactor_nrepl/find/find_symbol.clj
```diff
@@ -136,7 +137,12 @@
...
+                                                         (let [n (core/read-ns-form f)]
```
expez: Looks like this can throw, so should probably guard it with a `try/catch`

src/refactor_nrepl/core.clj
```diff
@@ -317,18 +317,27 @@
...
 
+(def require-lock (Object.))
+
+(defn safe-find-ns [n ignore-error?]
+  (try
+    (locking require-lock
+      (require n))
+    (find-ns n)
+    (catch Throwable e
+      (-> e .printStackTrace)
```
expez, line 328: Was this left here intentionally?

If so, does it make sense to change the `when-not` to an `if` so we don't print this when we're being asked to ignore errors?

src/refactor_nrepl/find/symbols_in_file.clj
```diff
@@ -56,8 +56,12 @@
   ([path parsed-ns dialect]
    (util/with-additional-ex-data [:file path]
      (let [dialect (or dialect (core/file->dialect path))
-           file-ns (or (find-ns (symbol (:ns parsed-ns))) *ns*)
-           ns-aliases (if (= dialect :cljs)
+           cljs? (= dialect :cljs)
+           file-ns (or (when-let [s (-> parsed-ns :ns symbol)]
+                         (when-not cljs?
+                           (core/safe-find-ns s true)))
```
expez, line 62: You don't have to change this, but in general it's nice to send in `:ignore-errors` instead of `true` as it makes the call-sites more readable when you don't have to remember how the truthy flags alter the behavior

src/refactor_nrepl/ns/slam/hound/search.clj
```diff
@@ -121,18 +121,18 @@
...
 (def available-classes-by-last-segment
-  (get-available-classes-by-last-segment))
+  (delay (get-available-classes-by-last-segment)))
 
 (defn reset
   "Reset the cache of classes"
   []
-  (alter-var-root #'available-classes (constantly (get-available-classes)))
-  (alter-var-root #'available-classes-by-last-segment (constantly (get-available-classes-by-last-segment))))
+  (alter-var-root #'available-classes (constantly (delay (get-available-classes))))
```
expez, line 137: Unless I'm reading it wrong you'll get a delay in a delay here (and on the line below)?

src/refactor_nrepl/analyzer.clj
```diff
@@ -91,7 +91,12 @@
     (when ns
       (if-let [cached-ast-or-err (get-ast-from-cache ns file-content)]
         cached-ast-or-err
-        (when-let [new-ast-or-err (try (build-ast ns aliases) (catch Throwable th th))]
+        (when-let [new-ast-or-err (try
+                                    (build-ast ns aliases)
+                                    (catch Throwable th
+                                      (when (System/getProperty "refactor-nrepl.internal.log-exceptions")
```
expez, line 97: Might be nice to pull this into a helper.  Not would would that read better but would make it easier to re-use without looking up the name of the property.

When doing 'information hiding' like this, I like to use the prefix `maybe-` for conditional actions like this.  `maybe-debug-log-exception`?  That makes code like this very easy to skip (if you don't care about the encapsulated details) during code reading later.

expez (changes requested):
Wow! 😃 Great work! 

Really appreciate the fine-grained commits, made review much easier.

Looks fantastic in general, so just some minor suggestions on my part.

vemv:
🍻 Nice catches! Addressed

expez (approved):
LGTM 👍

vemv:
Let's go 🚀

bbatsov:
Great work! 🎖️

bbatsov:
> Replace cheshire with data.json with a green build

Out of curiosity - which commit exactly fixed the failing test there?

vemv:
> Out of curiosity - which commit exactly fixed the failing test there?

Good q! I think it was the `test/` refactoring as I suspect the prior structure was confusing mranderson.

Ideally the PR would have been just that commit, but that refactoring surfaced some things that would have made the build fail (so the change would have been hard to defend).

bbatsov:
Well, I think it's safe to say no one will be complaining about the final outcome. :D

Issue #316, opened 2021-07-02
Review usages of `future-cancel`


vemv:
The couple current usages of `future-cancel` might not be accomplishing much if there isn't an active check for `(-> (Thread/currentThread) .isInterrupted)` in the corresponding threads doing the work.

PR #319, opened 2021-07-08
Observe thread interruption status for cancellable tasks

vemv:
Otherwise `future-cancel` wouldn't do much - at least from what I can infer reading the code.

Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/316

### DIFF
--- a/src/refactor_nrepl/ns/tracker.clj
+++ b/src/refactor_nrepl/ns/tracker.clj
@@ -55,7 +55,8 @@
   (let [deps (dep/immediate-dependents (:clojure.tools.namespace.track/deps tracker)
                                        (symbol my-ns))]
     (for [[file ns] (:clojure.tools.namespace.file/filemap tracker)
-          :when ((set deps) ns)]
+          :when ((set deps) ns)
+          :when (not (util/interrupted?))]
       file)))
 
 (defn- in-refresh-dirs? [refresh-dirs file]
--- a/src/refactor_nrepl/ns/tracker.clj
+++ b/src/refactor_nrepl/ns/tracker.clj
@@ -53,10 +53,11 @@
   "Get the dependent files for ns from tracker."
   [tracker my-ns]
   (let [deps (dep/immediate-dependents (:clojure.tools.namespace.track/deps tracker)
-                                       (symbol my-ns))]
+                                       (symbol my-ns))
+        deps-set (set deps)]
     (for [[file ns] (:clojure.tools.namespace.file/filemap tracker)
-          :when ((set deps) ns)
-          :when (not (util/interrupted?))]
+          :when (and (not (util/interrupted?))
+                     (deps-set ns))]
       file)))
 
 (defn- in-refresh-dirs? [refresh-dirs file]
--- a/src/refactor_nrepl/util.clj
+++ b/src/refactor_nrepl/util.clj
@@ -95,5 +95,7 @@
   Observing this condition helps `future-cancel` effectively cancel `future`s."
   ([]
    (interrupted? ::_))
-  ([_x]
+  ;; The arity with a "useless" arg is there so that this can be used as a predicate
+  ;; in other places that already are using `some-fn`, `every-pred`, etc
+  ([_]
    (.isInterrupted (Thread/currentThread))))
--- a/src/refactor_nrepl/util.clj
+++ b/src/refactor_nrepl/util.clj
@@ -88,3 +88,12 @@
           (maybe-log-exception e)
           ;; return false, because `with-suppressed-errors` is oriented for predicate usage
           false)))))
+
+(defn interrupted?
+  "Has the current thread been interrupted?
+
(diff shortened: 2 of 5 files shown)


expez:
> Otherwise future-cancel wouldn't do much

What do you mean by this?  The code will stop running because the JVM will raise an `InterruptedException`.  Unless this exception is caught to force work to continue (which I don't believe is the case) then all work in the thread will stop.  You can easily test this by dereferencing one of the futures we throw away.  If this is working properly an `InterruptedException` will surface.

I don't think there's any reason to make the code more complex since we don't have any resources that need to be cleaned up once we notice we're being interrupted.

expez:
FWIW I tested this quickly in the repl:

```clj
(def f
  (future
    (try (loop []
           (println "Running...")
           (Thread/sleep 1000)
           (recur))
         (catch Exception e
           (println e)
           (println  " interrupted? " (.isInterrupted (Thread/currentThread)))))))

(future-cancel f)
```

Will give this output:

```


Running...
#error {
        :cause sleep interrupted
        :via
        [{:type java.lang.InterruptedException
          :message sleep interrupted
          :at [java.lang.Thread sleep Thread.java -2]}]
        :trace
        [[java.lang.Thread sleep Thread.java -2]
         [hearing_aid.events$fn__37360 invokeStatic form-init1535742465164334437.clj 1599]
         [hearing_aid.events$fn__37360 invoke form-init1535742465164334437.clj 1596]
         [clojure.core$binding_conveyor_fn$fn__5754 invoke core.clj 2030]
         [clojure.lang.AFn call AFn.java 18]
         [java.util.concurrent.FutureTask run FutureTask.java 266]
         [java.util.concurrent.ThreadPoolExecutor runWorker ThreadPoolExecutor.java 1149]
         [java.util.concurrent.ThreadPoolExecutor$Worker run ThreadPoolExecutor.java 624]
         [java.lang.Thread run Thread.java 748]]}
interrupted?  false
```

The status of `interrupted` was the only thing that surprised me, but I suppose that makes sense when I think about it:  Futures aren't threads, so this is checking if the thread, in the thread pool, that's running the future is interrupted.

vemv:
> What do you mean by this? The code will stop running because the JVM will raise an InterruptedException.

That InterruptedException is not guaranteed to be thrown.

In general thread cancellation (that is, idiomatic cancellation that doesn't use the deprecated `Thread.stop` method) is presumed to be cooperative, not preemptive.

Here's a good read: https://codeahoy.com/java/Cancel-Tasks-In-Executors-Threads/

> [...] Will give this output:

That code perceives the exception because it has a explicit call to `Thread/sleep`. That's one of the very few safe "checkpoints" where one can preempt execution.

Otherwise execution cannot be preempted in arbitrary code - for very the same reason `Thread.stop` is deprecated.

Because this PR targets code that does not `sleep`, I'd say it remains valid.

expez:
> That code perceives the exception because it has a explicit call to Thread/sleep. That's one of the very few safe "checkpoints" where one can preempt execution.

Interesting!

I changed the `Thread/sleep` to this: `(reduce  + 0 (range))` and no longer got the interrupted exception but a REPL that needed to be restarted 🙄 

The java docs are very sparse on details.  Thanks for clarifying!

src/refactor_nrepl/ns/tracker.clj
```diff
@@ -55,7 +55,8 @@
   (let [deps (dep/immediate-dependents (:clojure.tools.namespace.track/deps tracker)
                                        (symbol my-ns))]
     (for [[file ns] (:clojure.tools.namespace.file/filemap tracker)
-          :when ((set deps) ns)]
+          :when ((set deps) ns)
+          :when (not (util/interrupted?))]
```
expez, line 59: Do we need this?  This final filtering of files using a set membership should be blazing fast.

src/refactor_nrepl/util.clj
```diff
@@ -88,3 +88,12 @@
...
           ;; return false, because `with-suppressed-errors` is oriented for predicate usage
           false)))))
+
+(defn interrupted?
+  "Has the current thread been interrupted?
+
+  Observing this condition helps `future-cancel` effectively cancel `future`s."
+  ([]
+   (interrupted? ::_))
+  ([_x]
```
expez, line 98: This looks a bit weird on its own.  I'd call the ignored param simply `_` and maybe even add a comment about having an extra arity for using in predicates.  Otherwise someone is likely to waste 2 minutes trying to see if they can refactor it.

expez (approved):
LGTM 👍

src/refactor_nrepl/ns/tracker.clj
vemv: Note that we're inside a `for` so `((set deps) ns)` will be invoked many times. So by observing cancellation we make the `for` interruptible.

Still, thanks for the scrutiny. I changed two things now:

* Calculate `(set deps)` just once
* Invoke `util/interrupted?` _prior_ to the set membership query - I'd expect the former to be faster.

Issue #230, opened 2018-07-20
cljr-slash is very slow


kommen:
## Expected behavior
Typing a `/` is fast

## Actual behavior
Typing a `/` makes emacs hang every time for about 0.5 to ~1 second.

## Steps to reproduce the problem

* have a significantly sized clojure project
* type `/` in clojure file

## Environment & Version information

spacemacs (current develop branch)

### clj-refactor.el and refactor-nrepl version information

clj-refactor 2.4.0-SNAPSHOT (package: 20180708.57), refactor-nrepl 2.4.0-SNAPSHOT

### CIDER version information

```
;; CIDER 0.18.0snapshot (package: 20180719.542), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_102
```

### Emacs version

GNU Emacs 26.1 (build 1, x86_64-apple-darwin17.5.0, NS appkit-1561.40 Version 10.13.4 (Build 17E199)) of 2018-05-28

### Operating system

macOS 10.13.6

### Profiler

```
- command-execute                                                1715  77%
 - call-interactively                                            1715  77%
  - funcall-interactively                                        1269  57%
   - cljr-slash                                                  1234  55%
    - cljr--magic-requires-lookup-alias                          1233  55%
     - cljr--get-aliases-from-middleware                         1233  55%
      - cljr--clj-context-p                                       762  34%
       - cljr--prompt-user-for                                    762  34%
        - completing-read                                         762  34%
         + #<compiled 0x50071a75>                                 762  34%
      - cljr--call-middleware-for-namespace-aliases                471  21%
       - cljr--call-middleware-sync                               354  15%
        + cider-nrepl-send-sync-request                           354  15%
       + edn-read                                                 117   5%
    + cljr--in-keyword-sans-alias-p                                 1   0%
```

expez:
Always or only the first time?  It builds up a cache the first time this command is called.  Parsing all namespaces in the project to find the most likely candidates can obv take some time i large projects.

kommen:
@expez just verified that in .clj files it is actually fast as expected, but in .cljs and .cljc files it is always slow.

expez:
Maybe the cache isn't working as it should for those dialects.  Thanks for investigating.

vemv:
I wonder if addressing this one would be a matter of parallelising the workload (hypothesis: cljs workloads tend to be larger for some reason, hence the performance difference)

[...]

PR #320, opened 2021-07-08
Parallelize `referred-syms-by-file&fullname`

vemv:
Part of https://github.com/clojure-emacs/refactor-nrepl/issues/230

Parallelizes `referred-syms-by-file&fullname` and `core/` functions used by `referred-syms-by-file&fullname`.

(the `core/` parallelization will be only noticable in large projects)

One can verify the improvement with the following repl input over refactor-nrepl itself:

```clj
(require 'refactor-nrepl.ns.libspecs)
(in-ns 'refactor-nrepl.ns.libspecs)
(time (referred-syms-by-file&fullname true))
```

...It's clearly better than `master` performance for both the non-cached and cached (i.e. subsequent invocations) cases.

(the numbers of course will depend on your CPU count)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 #### Changes
+* [(Part of #230)](https://github.com/clojure-emacs/refactor-nrepl/issues/230) Parallelize various fuctionality
+  * This will have a noticeable improvement in e.g. clj-refactor.el's `cljr-slash` performance.
 * [#291](https://github.com/clojure-emacs/refactor-nrepl/issues/291): The `:ignore-errors` option will be honored in more places, making refactor-nrepl more robust in face of files not particularly meant to be part of the AST corpus.
   * Examples: WIP files, Moustache template files, scripts.
 * Upgrade Orchard
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## Unreleased
 
 #### Changes
-* [(Part of #230)](https://github.com/clojure-emacs/refactor-nrepl/issues/230) Parallelize various fuctionality
+
+* [(Part of #230)](https://github.com/clojure-emacs/refactor-nrepl/issues/230): Parallelize various functionality
   * This will have a noticeable improvement in e.g. clj-refactor.el's `cljr-slash` performance.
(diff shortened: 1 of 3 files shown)


src/refactor_nrepl/core.clj
```diff
@@ -184,7 +188,10 @@
 (defn find-in-project
   "Return the files in the project satisfying (pred ^File file)."
   [pred]
-  (-> find-in-dir (partial pred) (mapcat (dirs-on-classpath)) distinct))
+  (->> (dirs-on-classpath)
+       (pmap (partial find-in-dir pred))
```
vemv, line 192: I chose `pmap` for simplicity, it's better than some tend to say

(because `pmap` won't create one thread per item... it takes the CPU count in account)

...it would be inefficient for very small inputs, but it's safe to say that most projects have multiple dirs in the classpath (especially in monorepos), plenty of namespaces, etc.

src/refactor_nrepl/ns/libspecs.clj
```diff
@@ -88,10 +88,10 @@
    {:clj  (->> (core/find-in-project (util/with-suppressed-errors
                                        (some-fn core/clj-file? core/cljc-file?)
                                        ignore-errors?))
-               (map (juxt identity (partial get-libspec-from-file-with-caching :clj)))
+               (pmap (juxt identity (partial get-libspec-from-file-with-caching :clj)))
```
bbatsov, line 91: I'd add a note about the use of `pmap`, as people tend to "optimize" it by removing it. :D

bbatsov (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased
 
 #### Changes
+* [(Part of #230)](https://github.com/clojure-emacs/refactor-nrepl/issues/230) Parallelize various fuctionality
```
bbatsov, line 6: One blank line is missing. And a `:`.

src/refactor_nrepl/core.clj
expez: Good choice! The small workloads that will get slightly slower will still be more than fast enough in the face of some fork/join overhead, I think  :)

expez (approved):
LGTM 👍

vemv:
🍻

Issue #245, opened 2019-03-28
find-symbol can fail with "Requested array size exceeds VM limit"


vemv:
Hi there,

I've faced this issue repeatedly. In the past I had worked around it via `:jvm-opts ["-Xmx18G"]`, but today I hit it again (long time no see!). I could try setting an even higher bound, but I guess a GH issue will be more useful.

The error is:

* When I cljr-rename-symbol, through natural usage,
* I get the attached stacktrace printed to Emacs' `*Messages*` 

```
Fetching symbol occurrences...
Error in nrepl-refactor: java.lang.OutOfMemoryError: Requested array size exceeds VM limit
 at java.util.Arrays.copyOf (Arrays.java:3332)
    java.lang.AbstractStringBuilder.ensureCapacityInternal (AbstractStringBuilder.java:124)
    java.lang.AbstractStringBuilder.append (AbstractStringBuilder.java:448)
    java.lang.StringBuffer.append (StringBuffer.java:270)
    java.io.StringWriter.write (StringWriter.java:101)
    clojure.core$fn__7283.invokeStatic (core_print.clj:159)
    clojure.core/fn (core_print.clj:158)
    clojure.lang.MultiFn.invoke (MultiFn.java:234)
    clojure.core$pr_on.invokeStatic (core.clj:3674)
    clojure.core$pr_on.invoke (core.clj:3668)
    clojure.core$print_prefix_map$fn__7317.invoke (core_print.clj:233)
    clojure.core$print_sequential.invokeStatic (core_print.clj:66)
    clojure.core$print_prefix_map.invokeStatic (core_print.clj:229)
    clojure.core$print_map.invokeStatic (core_print.clj:238)
    clojure.core$fn__7344.invokeStatic (core_print.clj:266)
    clojure.core/fn (core_print.clj:263)
    clojure.lang.MultiFn.invoke (MultiFn.java:234)
    clojure.core$pr_on.invokeStatic (core.clj:3674)
    clojure.core$pr_on.invoke (core.clj:3668)
    clojure.core$print_prefix_map$fn__7317.invoke (core_print.clj:233)
    clojure.core$print_sequential.invokeStatic (core_print.clj:66)
    clojure.core$print_prefix_map.invokeStatic (core_print.clj:229)
    clojure.core$print_map.invokeStatic (core_print.clj:238)
    clojure.core$fn__7344.invokeStatic (core_print.clj:266)
    clojure.core/fn (core_print.clj:263)
[...]

PR #321, opened 2021-07-09
Always perform a `require` prior to analyzing a ns

vemv:
Part of https://github.com/clojure-emacs/refactor-nrepl/issues/245

Performing this `require` makes it more likely that t.ana will succeed. I believe it's becase `require` will also `require` other stuff recursively. t.ana does so in theory as well but it's a bit more rigid.

### DIFF
--- a/src/refactor_nrepl/analyzer.clj
+++ b/src/refactor_nrepl/analyzer.clj
@@ -78,7 +78,12 @@
     ;; Use `locking`, because AST analysis can perform arbitrary evaluation.
     ;; Parallel analysis is not safe, especially as it can perform `require` calls.
     (locking core/require-lock ;; for both `require` and `aj/analyze-ns`
-      (require ns) ;; ease the work for t.ana
+
+      ;; Performing this `require` makes it more likely that t.ana will succeed.
+      ;; I believe it's because `require` will also require other namespaces recursively.
+      ;; t.ana does so in theory as well, but it's slightly more rigid,
+      ;; and/or does not always do the same exact thing the Clojure compiler would.
+      (require ns)
       (let [opts {:passes-opts
                   {:validate/unresolvable-symbol-handler shadow-unresolvable-symbol-handler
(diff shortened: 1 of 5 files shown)


test/refactor_nrepl/analyzer_test.clj
```diff
@@ -0,0 +1,10 @@
+(ns refactor-nrepl.analyzer-test
+  (:require
+   [clojure.java.io :as io]
+   [refactor-nrepl.analyzer :as sut]
+   [clojure.test :refer [deftest is]]))
+
+(deftest ns-ast-test
```
vemv, line 7: I verified that if we comment out this PR's main change, this test would fail, complaining about `go-loop`

src/refactor_nrepl/analyzer.clj
```diff
@@ -77,7 +77,8 @@
              (not (util/self-referential? ns)))
     ;; Use `locking`, because AST analysis can perform arbitrary evaluation.
     ;; Parallel analysis is not safe, especially as it can perform `require` calls.
-    (locking core/require-lock
+    (locking core/require-lock ;; for both `require` and `aj/analyze-ns`
+      (require ns) ;; ease the work for t.ana
```
expez, line 81: Here's a general tip with regards to writing comments: have them answer *why questions*.

Since it's here we can assume that it's needed, but we're still left asking why that is.  You could remedy that with a comment like:  `;; t.ana sometimes fails to require the needed namespaces on its own`

expez (approved):
Man, you are on a tear 🥇 

Great job!

src/refactor_nrepl/analyzer.clj
bbatsov: I'd use this as the comment here:

```
Performing this require makes it more likely that t.ana will succeed. I believe it's becase require will also require other stuff recursively. t.ana does so in theory as well but it's a bit more rigid.
```

vemv:
Thanks for the suggestions!

PR #324, opened 2021-07-18
Address remaining reflection warnings

vemv:


project.clj
```diff
@@ -46,8 +46,7 @@
 
              :lein-plugin {:source-paths ["lein-plugin"]}
              :test {:dependencies [[print-foo "1.0.2"]]}
-             :dev {:global-vars {*warn-on-reflection* true}
```
vemv, line 49: Not needed anymore. Eastwood 0.9+ lints for reflection warnings, regardless of any per-project/per-ns `*warn-on-reflection*` setup.

This approach is nicer because only reflection warns coming from our own source-paths / test-paths will cause a CI failure. Other warns won't even show up in stdout, resulting in less noise

expez (approved):
💯

vemv:
🍻!

PR #326, opened 2021-09-13
Misc / minor

vemv:
In preparation to another PR.

* Upgrade and satisfy Eastwood
  * Mainly satisfies `:boxed-math`
  * To be clear, it doesn't necessarily always result in a perf gain; it can merely make the underlying boxing more explicit.
* Tests: handle .nrepl-port files with newlines in them
* Make `artifacts-file` cross-OS
* Remove `-Djava.net.preferIPv4Stack=true`
  * The build passes without it, and it was preventing a repl from being successfully launched in my environment.
* Use `require-lock` in the last remaining place

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -66,12 +66,13 @@
                                           with-debug-bindings [[:inner 0]]
                                           merge-meta [[:inner 0]]
                                           try-if-let [[:block 1]]}}}]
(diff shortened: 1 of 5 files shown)


bbatsov (approved):
Great work!

PR #327, opened 2021-09-16
Avoid a lazy `pmap`ping

vemv:
Ensures `pmap` work as intended

I verified that no other places needed a fixing.

### DIFF
--- a/src/refactor_nrepl/core.clj
+++ b/src/refactor_nrepl/core.clj
@@ -102,6 +102,8 @@
   Note that files which are non-existent, hidden or build-artifacts
   are pruned by this function."
   [pred dir]
+  ;; ensure pmap is not used lazily:
+  {:post [(vector? %)]}
   (->> dir
        file-seq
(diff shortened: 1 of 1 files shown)


bbatsov (approved):

bbatsov:
Thanks!

expez:
Why is this necessary?  If the value this function produces is needed then this change makes no difference, right?  Conversely, if the value is never needed then this is a regression?

Asking for a friend who's a fairly naive `pmap` user 😄

vemv:
The key is in how the touched defn is used: https://github.com/clojure-emacs/refactor-nrepl/blob/eb77344780482e4d9a80c4b6124a7d3937055133/src/refactor_nrepl/core.clj#L199

it's essentially a `(pmap (pmap ...`. If the inner pmap is lazy, then the outer pmap isn't doing something much useful i.e. we're creating a lazy list in parallel.

The final lazy seq consumer would do all the work, outside any pmap

Issue #288, opened 2020-07-23
[client-loop] ERROR - select exception, should not happen  Client/Server mode has not yet been set


nfedyashev:
## Expected behavior
I would like to use cljr to update all the project dependencies like I did in other clj repos.

## Actual behavior
Checking/updating project.clj versions work fine until it reaches the line with
[clojure-csv/clojure-csv "2.0.1"]

Then it causes 100% of CPU usage on one core, freezed emacs and in ~20 seconds  the following error:

```
Fri Jul 23 16:23:35 EEST 2020 [client-loop] ERROR - select exception, should not happen
java.lang.IllegalStateException: Client/Server mode has not yet been set.
	at java.base/sun.security.ssl.SSLEngineImpl.beginHandshake(SSLEngineImpl.java:98)
	at refactornrepl250SNAPSHOT.org.httpkit.client.HttpClient.finishConnect(HttpClient.java:395)
	at refactornrepl250SNAPSHOT.org.httpkit.client.HttpClient.run(HttpClient.java:472)
	at java.base/java.lang.Thread.run(Thread.java:834)
```
As a workaround I can comment out this dependency(temporary) and run cljr-update-project-dependencies and it works fine.
I've tried to completely restart the emacs - this issue could be reproduced reliably.

## Steps to reproduce the problem

a project.clj with 
[clojure-csv/clojure-csv "2.0.1"] somewhere in the middle
and run cljr-update-project-dependencies.

## Environment & Version information

clj-refactor.el 20200405.1419
[refactor-nrepl "2.5.0"]

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; CIDER 0.12.0snapshot (package: 20160331.421), nREPL 0.2.12
;; Clojure 1.8.0, Java 1.8.0_31
```

### Leiningen or Boot version

Leiningen 2.9.4 on Java 11.0.6 OpenJDK 64-Bit Server VM

### Emacs version

GNU Emacs 26.1 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.5) of 2019-09-23, modified by Debian

### Operating system

MXLinux/Debian

vemv:
This was fixed (intentionally or not 😄) with https://github.com/clojure-emacs/refactor-nrepl/commit/89d678c7c5dd261fc59b9c29aac98a57814e361e and released back then as refactor-nrepl 2.5.0

I'll be adding some integration tests anyway

PR #329, opened 2021-09-17
Improve http-kit integration / test coverage

vemv:
See also: https://github.com/clojure-emacs/refactor-nrepl/issues/288

### DIFF
--- a/src/refactor_nrepl/artifacts.clj
+++ b/src/refactor_nrepl/artifacts.clj
@@ -51,7 +51,7 @@
          ;; Ignore artifact if not readable. See #255
          nil)))
 
-(defn get-clojars-artifacts!
+(defn- get-clojars-artifacts!
   "Returns a vector of [[some/lib \"0.1\"]...]."
   []
   (try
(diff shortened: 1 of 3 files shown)


src/refactor_nrepl/artifacts.clj
```diff
@@ -51,35 +51,42 @@
...
   "Returns a vector of [[some/lib \"0.1\"]...]."
   []
   (try
-    (->> "https://clojars.org/repo/all-jars.clj.gz"
-         io/input-stream
-         GZIPInputStream.
-         io/reader
-         line-seq
-         (keep edn-read-or-nil))
+    (or @clojars-artifacts-cache
```
expez, line 60: I believe this breaks the `force` functionality in `artifact-list` which forces an update of a potentially stale cache.

expez (commented):
Looks good otherwise :)

src/refactor_nrepl/artifacts.clj
vemv: Mmm good once, I briefly considered the daily caching but it didn't seem too relevant.

It is for long-lived sessions which are very much emacsy :)

vemv:
Removed the atom cache I had added and made that one defn private instead. That way one can only use said defn via its caching façade.

expez (approved):
LGTM 👍

PR #330, opened 2021-09-18
Fix `in-refresh-dirs?`

vemv:
`#(str/starts-with? % file-as-absolute-paths)` was invoking the arguments in the wrong order

As a postmortem, sometimes we use defn privacy as a hatch for skipping the authoring of deftests, here it turned out to be too much of an important feature to skip coverage.

While I was there, I also optimized things as bit by absolutizing the refresh-dirs names less redundantly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 VERSION ?= 1.10
 
 .inline-deps:
+	git clean -fdx
 	lein with-profile -user,+$(VERSION) inline-deps
 	touch .inline-deps
 
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,11 @@
 
 VERSION ?= 1.10
 
-.inline-deps:
-	git clean -fdx
+clean:
+	lein clean
+	rm -f .inline-deps
+
+.inline-deps: clean
 	lein with-profile -user,+$(VERSION) inline-deps
 	touch .inline-deps
 
@@ -37,6 +40,5 @@ release:
 deploy: .inline-deps
 	lein with-profile -user,+$(VERSION),+plugin.mranderson/config,+lein-plugin deploy clojars
 
-clean:
-	lein clean
-	rm -f .inline-deps
+install: .inline-deps
+	lein with-profile -user,+$(VERSION),+plugin.mranderson/config,+lein-plugin install
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject refactor-nrepl "3.0.0-alpha1"
+(defproject refactor-nrepl "3.0.0-alpha4"
   :description "nREPL middleware to support editor-agnostic refactoring"
   :url "http://github.com/clojure-emacs/refactor-nrepl"
   :license {:name "Eclipse Public License"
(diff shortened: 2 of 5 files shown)


vemv:
For some reason `make deploy` can publish empty or semi-empty releases...

Makefile
```diff
@@ -3,6 +3,7 @@
 VERSION ?= 1.10
 
 .inline-deps:
+	git clean -fdx
```
vemv, line 6: This appeared to fix the mrandersoning, probably it'd be better to improve the `clean` task

vemv (commented):
PR ready, sorry for the excess emails

src/refactor_nrepl/ns/tracker.clj
```diff
@@ -60,31 +60,38 @@
                      (deps-set ns))]
       file)))
 
-(defn- in-refresh-dirs? [refresh-dirs file]
+(defn- absolutize-refresh-dirs [refresh-dirs]
```
bbatsov, line 63: Won't it be better to have a generic helper function operating on dirs and just apply it to `refresh-dirs`? Seems cleaner to me.

project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject refactor-nrepl "3.0.0-alpha4"
+(defproject refactor-nrepl "3.0.0-alpha10"
```
bbatsov, line 1: Why the alpha versions keep changing in every commit?

src/refactor_nrepl/ns/tracker.clj
```diff
@@ -4,7 +4,7 @@
             [clojure.tools.namespace
              [dependency :as dep]
              [file :as file]
-             [repl :refer [refresh-dirs]]
+             [repl :as tools.namespace.repl]
```
bbatsov, line 7: That's inconsistent with the rest of the aliases. I kind of understand the reasoning, but I don't like inconsistencies, especially without explanations. :-)

vemv: `repl` in particular has bit of an unfortunate name, in principle the `refresh-dirs` aren't exclusively concerned with the repl.

When using refactor-nrepl one might not interact with a repl at all to get certain stuff done.

The `tools.namespace.repl` alias moves weight off 'repl' and into 'tools.namespace'.

project.clj
vemv: > Why the alpha versions keep changing in every commit?

I had shared the alpha with someone and found out that we had a bunch of buggy stuff + buggy Makefile 😞 so I didn't stop until I got something decent shared publicly.

I added a `make install` command in this PR so I won't need alphas as much anyway!

src/refactor_nrepl/ns/tracker.clj
bbatsov: Yeah, I got this. I meant it probably won't be clear for everyone. I'd just adopt consistent aliases like `c.t.n.repl`, `c.t.n.file`, etc. Or something along those lines at least.

vemv: Hmmm, I like / I'm used to `file` as an alias from various projects. 

Maybe I can set up `:consistent-alias` in clj-kondo for enforcing that we're consistently inconsistent :) i.e. just one exception to these short names.

I'd add an explanatory comment as well to that linting config. SGTY?

vemv: > I'd add an explanatory comment as well to that linting config

(done, tentatively)

bbatsov (approved):

PR #331, opened 2021-09-18
Restore tools.nrepl compatibility

vemv:
Blast from the past...

Personally I keep using tools.nrepl, I want to stop doing that but can't atm.

These changes help me try the refactor-nrepl alphas before they hit a broader public.

This commit reverts 345fd20 with an additional system property for extra caution.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,4 +1,5 @@
 ;; `nrepl.core/code` can contain intentionally broken code in the test suite, so we silence related resolution errors:
 {:lint-as {nrepl.core/code clojure.core/quote}
- :linters {:unresolved-symbol {:exclude [(refactor-nrepl.ns.ns-parser/with-libspecs-from [libspecs])]}
+ :linters {:unresolved-symbol {:exclude [(refactor-nrepl.ns.ns-parser/with-libspecs-from [libspecs])
+                                         (refactor-nrepl.middleware/set-descriptor! [set-descriptor!])]}
            :unresolved-namespace {:exclude [clojure.main]}}}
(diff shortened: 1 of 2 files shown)


bbatsov:
> Personally I keep using tools.nrepl, I want to stop doing that but can't atm.

Can you remind me why do you need to use `tools.nrepl`? I'm very wary of this change, as it makes it much harder for people who are not very familiar with tooling infrastructure to understand why something is broken for them (e.g. they have some old `tools.nrepl` dep that gets picked up and they wonder why something is not behaving as expected). This was the main reason I wanted to get rid of the legacy compatibility - I got a ton of the issues/support requests that were essentially more of the same.

vemv:
> Can you remind me why do you need to use tools.nrepl? 

It mainly has to do with the fact that I froze / forked cider.el a few years ago for personal consumption. I'm trying to get rid of that but it's not trivial.

> This was the main reason I wanted to get rid of the legacy compatibility

I took this into account and hid the feature under a System property.

bbatsov:
Got it. This was the other part of my argument - the current CIDER won't work properly with the old `tools.nrepl` anyways (e.g. because some bugs were fixed downstream and we removed workarounds in CIDER and so on). Anyways, I was going to suggest a flag if we really want to do this and it seems you came to the same conclusion yourself. I'd suggest a simpler name for the flag like `use/try-tools.nrepl`, but I don't really care about the name.

vemv:
Luckily enough refactor-nrepl latest is working fine with the old tools.nrepl. There was a reasonable chance it wouldn't 😄

(I have an old / forked cider-nrepl in my stack as well)

I like the sysprop name, it's both informative (tells you the project it relates to) and scary for outsiders

bbatsov (approved):
Fair enough.

Issue #335, opened 2021-09-20
Optimize `resolve-missing`


vemv:
https://github.com/clojure-emacs/refactor-nrepl/commit/1399da2029c2e875d450af4caa347302f55fa965 worked nicely for `namespace-aliases`,  I definitely perceived an improvement in `cljr-slash` perf in my day job codebase.

Now `resolve-missing` is another of the big features offered by refactor-nrepl and deserves a similar treatment.

`refactor-nrepl.ns.slam.hound.regrow` is the relevant ns.

## Notes

* https://github.com/clojure-emacs/orchard/issues/125 is a blocker
* there's a double `delay`
  * this means some missing test coverage
* the `warm-ast-cache` op could `@` all delays, for improving performance

PR #336, opened 2021-09-25
Strengthen `resolve-missing`

vemv:
Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/335

* Remove `slam.hound.future`
  * The functions provided by it are expected to be present in all end users' environments.
  * (Especially since rewrite-clj now has a min requirement of Clojure 1.8)
* Multiple fixes for `refactor-nrepl.ns.slam.hound.search`
  * Handle `delay` correctly
  * Handle separators correctly for Windows
  * Use a classpath value that will work
  * Filter out more Clojure auto-generated classes
  * Remove 'namespace' finding functionality, we don't use it (as we use `all-ns` instead) so it was confusing.
* Remove unused code from `refactor-nrepl.ns.slam.hound.regrow`


bbatsov:
Good changes overall, but I guess it's time to drop this code structure with the copied slamhound namespaces and just put everything new need under `refactor-nrepl.ns` or something like this. slamhound is dead and buried at this point and the current code structure is confusing for people not familiar with how we got where we are. A good opportunity to simplify things. Perhaps down the road some of this stuff can go to `orchard`? We'll see.

vemv:
Good remark, I think that over time I'll be moving/proposing new functionality to Orchard so that it fulfills its original purpose optimally.

vemv:
Ready again

bbatsov (approved):

PR #338, opened 2021-10-03
`imports-and-refers-analysis`: remove tracking and caching

vemv:
Tracking was excessively obstrusive (it monkey-patched clojure.repl) and coupled (to nrepl middleware).

Now `symbols->ns-syms` doesn't use caching at all, as it ran in less than 1ms anyway.

Similarly `all-ns-imports` has been removed of any caching. This function was a little more expensive (100ms for a medium-sized project), but parallelizing it brings that cost down to 20ms which seems more than acceptable for an end user.

These changes are backed by `refactor-nrepl.ns.imports-and-refers-analysis-test`.


expez (approved):
Great work! :100:

vemv:
🍻!

PR #340, opened 2021-10-05
Make `refactor-nrepl.plugin` available again.

vemv:
Copies `refactor-nrepl.plugin` as-is from `lein-plugin/refactor_nrepl` to `src/refactor_nrepl`

It turned out, mranderson seemingly supports only one entry for `:source-paths`, so extra entries would be silently dropped, making the plugin unavailable in .jar releases.

I've successfully released this as https://clojars.org/refactor-nrepl/versions/3.0.0-alpha13 and got a confirmation that it works for the plugin use case.

Will keep the PR open until I find a solution other than copying the ns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Be aware that this isn't the case if you connect to an already running REPL proc
 Add the following, either in your project's `project.clj`,  or in the `:user` profile found at `~/.lein/profiles.clj`:
 
 ```clojure
-:plugins [[refactor-nrepl "3.0.0-alpha12"]
+:plugins [[refactor-nrepl "3.0.0-alpha13"]
           [cider/cider-nrepl "0.25.9"]]
 ```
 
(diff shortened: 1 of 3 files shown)


expez:
Linter seems to fail on the missing leiningen dep.  

I think we solved this last time by putting `:eval-in-leiningen true` in `project.clj`

vemv:
Hey there! Thanks for response, this is important stuff no doubt.

The separate `:source-paths` entry has worked well for Eastwood, as an example of another lein plugin that can decouple (and later compose) the plugin and non-plugin parts.

This has various benefits while developing/linting/testing a given project.

The difference here is that mranderson is being used. I'll create an issue upstream. From what an eyed, making it support more `:source-paths` should be at hand.

> Btw, I wonder why no one noticed the problem until now.

I didn't start to give (half-) serious sharing to the `3.*` series until last weekend. Personally I use either deps.edn or Lein in a non-plugin way so I had no way to catch this 😞

vemv:
I created https://github.com/benedekfazekas/mranderson/issues/58 . It seems at hand, anyway for the time being I'll just go the 'easy' way and one only one source dir.

Added a couple integration tests from different perspectives (`lein repl`, `lein jar`).

Review appreciated!

Issue #344, opened 2021-10-25
Don't avoid 1-element vectors/lists


vemv:
https://github.com/clojure-emacs/refactor-nrepl/pull/308 started going more into the [how to ns](https://stuartsierra.com/2016/08/27/how-to-ns) direction, however there's still a difference:

![image](https://user-images.githubusercontent.com/1162994/138682080-1b59635a-9d0f-4371-8d7d-902a6b6e383d.png)

I'd say we should favor lists/vectors (as opposed to unwrapping elements) so that:

* we don't contribute to a proliferation of styles
* things are sortable via `M-x sort-lines`
  * else it won't meaningfully sort in mixed presence of colls vs. symbols.
* indentation is consistent.

This particularly matters because e.g. `rename-file-or-dir` always performs a clean-ns, so renaming a ns should not change other aspects of its ns form (presumably formatted/linted with a variety of tools).

camsaul:
Is there some way to disable this change so we can keep the old functionality?

vemv:
We can introduce a feature flag (which would have to be settable from clj-refactor.el also), PRs welcome as well

btw would appreciate https://github.com/camsaul/whitespace-linter/issues/9 - might be a fair exchange 😇

PR #345, opened 2021-10-25
Don't avoid 1-element vectors/lists

vemv:
Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/344 / rationale there

Not necessarily for merging right away, before I'd like to know if you agree this is a better default?

Should we also provide an option to fall back?

Personally I wouldn't, there doesn't seem to be a huge rationale for the old behavior other than "make things shorter" which seems very relative when considering consistency with other tools or trends over the last few years.


bbatsov (approved):

camsaul:
Is there a way to disable this?

PR #348, opened 2021-11-04
Use source-dirs-on-classpath in more places

### LINKED ISSUE #333: Use `source-dirs-on-classpath` in more places
The source paths inference provided by the new `source-dirs-on-classpath` defn is much handy and can increase perf / accuracy.

* Use it in missing places
* Use it as a fallback if `refresh-dirs` are unset / empty

vemv:
Fixes #333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Changes
+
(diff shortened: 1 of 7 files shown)


bbatsov:
You might want to update your commit message. :-)

bbatsov (approved):

vemv:
Thanks for the review / heads up!

(will fix via squashing which makes the msg bearable)

PR #349, opened 2021-11-04
Refinements for `clean-ns`

vemv:
* Add `fast-test` Makefile task
* Decouple `ns.pprint` from `clojure.pprint` defaults
  * The previous behavior was to wrap around 80 colums. Often teams have a quite more generous column limit (if placing a limit at all), so it's simplest to remain agnostic.
  * An option is introduced for reverting to the previous behavior.
* `ns.rebuild`: also wrap single-segment namespace in a vector
  * An edge case I found out about just now.

### DIFF
--- a/src/refactor_nrepl/config.clj
+++ b/src/refactor_nrepl/config.clj
@@ -34,13 +34,24 @@
    :libspec-whitelist ["^cljsjs"]
 
    ;; Regexes matching paths that are to be ignored
-   :ignore-paths []})
+   :ignore-paths []
+
+   ;; Will be forwarded to clojure.pprint/*print-right-margin* when pprinting ns forms.
+   ;; You can set it to nil for disabling line wrapping.
+   ;; See also: :print-miser-width
+   :print-right-margin 72
+
+   ;; Will be forwarded to clojure.pprint/*print-miser-width* when pprinting ns forms.
+   ;; You can set it to nil for disabling line wrapping.
+   ;; See also: :print-right-margin
+   :print-miser-width 40})
 
 (defn opts-from-msg [msg]
   (into {}
         (map (fn [[k v]] (cond
                            (and (string? v) (= v "true")) [k true]
                            (and (string? v) (= v "false")) [k false]
+                           (and (string? v) (= v "nil")) [k nil]
                            :else [k v]))
              (update (select-keys msg (keys *config*))
                      :ignore-paths
--- a/src/refactor_nrepl/ns/pprint.clj
+++ b/src/refactor_nrepl/ns/pprint.clj
@@ -1,7 +1,7 @@
 (ns refactor-nrepl.ns.pprint
   (:require [cljfmt.core :as fmt]
             [clojure
-             [pprint :refer [pprint]]
+             [pprint]
              [string :as str]]
             [refactor-nrepl
              [config :refer [*config*]]
@@ -9,6 +9,16 @@
              [util :as util :refer [replace-last]]])
 
   (:import java.util.regex.Pattern))
+
+(defn pprint
+  "Pretty-prints `x` with custom, configurable overrides over `clojure.pprint`'s settings.
+
+  This way, we try to generate a formatting that is agnostic / decoupled from clojure.pprint,
+  and therefore that other tools can also consistently achieve."
+  [x]
+  (binding [clojure.pprint/*print-miser-width* (:print-miser-width *config*)
+            clojure.pprint/*print-right-margin* (:print-right-margin *config*)]
+    (clojure.pprint/pprint x)))
 
 (defn- libspec-vectors-last [libspecs]
   (vec (concat (remove sequential? libspecs)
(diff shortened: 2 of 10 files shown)


src/refactor_nrepl/ns/pprint.clj
```diff
@@ -10,6 +10,11 @@
 
   (:import java.util.regex.Pattern))
 
+(defn pprint [x]
```
bbatsov, line 13: Public functions need docstrings, even if they are as simple as this one.

src/refactor_nrepl/config.clj
```diff
@@ -34,7 +34,14 @@
    :libspec-whitelist ["^cljsjs"]
 
    ;; Regexes matching paths that are to be ignored
-   :ignore-paths []})
+   :ignore-paths []
+
+   ;; Will be forwarded to clojure.pprint/*print-miser-width* when pprinting ns forms.
+   ;; You can set it to 40 for restoring traditional behavior.
+   :print-miser-width nil
```
bbatsov, line 41: I guess this should also be documented in the README.

bbatsov (approved):
Overall the proposed changes seem fine to me.

src/refactor_nrepl/config.clj
```diff
@@ -34,7 +34,14 @@
...
+   ;; Will be forwarded to clojure.pprint/*print-right-margin* when pprinting ns forms.
+   ;; You can set it to 72 for restoring traditional behavior.
+   :print-right-margin nil})
```
expez, line 44: Java class names are sometimes very long.  If I'm not mistaken, the proposed change would mean that we could potentially create import clauses that would be so long nobody would be happy with them.

I don't think anyone would want that, even if I can see the argument that 72 is too restrictive.

Can we set this to 100 instead?  Having sensible defaults is very important imo, so the behavior you get out of the box is at least workable.

src/refactor_nrepl/config.clj
expez: Should probably mention this default change in the changelog too, like you did for `print-right-margin`

src/refactor_nrepl/config.clj
vemv: The tradeoff I'm making (which reasoning I only partly shared in advance - my bad!) is that I'm fostering an agnostic/minimalistic formatting that other tools can independently, consistently achieve.

In specific terms, our impl is based on clojure.pprint. Other similar ns formatters are using rewrite-clj or vanilla clojure.

So by coupling our result to a clojure.pprint impl detail, we create a non-reproducible formatting for other tools.

`how-to-ns`, the old "standard" for this domain doesn't mention a limit of columns. So it is cautious to stick to it without introducing extra rules.

One could always add the extra newlines either with the provided new config options, or perhaps a generalized "column formatter" e.g. zprint, i.e. teams who care about this most likely already have a formatter/linter for the entire problem (.clj files, beyond ns forms).

vemv:
Rebased w/ PR feedback

src/refactor_nrepl/config.clj
expez: > The tradeoff I'm making (which reasoning I only partly shared in advance - my bad!) is that I'm fostering an agnostic/minimalistic formatting that other tools can independently, consistently achieve.

That's fine, but making the tool harder to use / have worse formatting out of the box seems like a high price to pay.  It's essentially a race to the bottom, because the easiest and most consistent way to format the ns form is to put everything on a single line that runs on for however many columns is needed...

I was going to propose moving the (imo sensible) defaults to clj-refactor instead, but that wouldn't help, right?  What you want is essentially less clever formatting, so it can be widely replicated across the tooling universe?

When I wrote `clean-ns` back in the day, what I wanted most of all was to simply pull in a library that pretty printed the ns form in a nice way.  I looked at every alternative available at the time, but couldn't find one that worked well in practice.  Perhaps your goal could be even better served if we pulled this pretty printer out into it's own library so everyone else could re-use it (with whatever config settings they want)?

> how-to-ns, the old "standard" for this domain doesn't mention a limit of columns. So it is cautious to stick to it without introducing extra rules.

FWIW, if people were consistently submitting PRs to him with import clauses that were 140 columns long, when doing java interop, I'm sure he would've mentioned it, but nobody would do that when they were manually editing the ns form.  Tooling with bad defaults might, though!

meta comment: in the grand scheme of things this isn't a big deal and I can obviously live with whatever the outcome is, but since you opened a PR I figured I would chime in and fight for ease-of-use on behalf of our users :)  Writing this all out also made me think of this old xkcd strip:

vemv: I'd disagree about what the best formatting is anyway, historically refactor-nrepl's line wrapping meant that `M-x sort-lines` wouldn't work.

But keeping things practical, we can observe that this is a flip-coin scenario so we can default to not changing things. Setting .dir-locals is easy enough.

I also particularly care about keeping our own team happy 🍻

vemv:
Amended with a more conservative setting

PR #352, opened 2021-11-07
Make `ns.class-search` more comprehensive

vemv:
Finds more classes (e.g. `FileChannel`), in all JDKs, without needing an extra .jar (like rt.jar, which I was tempted to add for JDK8 but it's a pretty dubious thing to do) in the classpath to be present.

For this we use Compliment's excellent impl which takes care about cross-JDK compat, and some edge cases e.g. Android: https://github.com/alexander-yakushev/compliment/blob/0e0d3e68c665848522d0b6d1f966920dd10ed7ca/src/compliment/utils.clj

Between manual testing and `refactor-nrepl.ns.class-search-test` I'm confident about these changes.

Finally, while Compliment might not be exactly a class searching API, by mrandersoning this dep we can be sure that this feature will never go away.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Changes
 
 * [#344](https://github.com/clojure-emacs/refactor-nrepl/issues/344): make clean-ns's style closer to the [how to ns](https://stuartsierra.com/2016/08/27/how-to-ns) style.
-* [#333](https://github.com/clojure-emacs/refactor-nrepl/issues/333): skip scanning irrelevant directories in more places (as it was already done for various other functionalities; this limits ananlysis/refactoring to your source/test paths, skipping other artifacts). 
+* [#333](https://github.com/clojure-emacs/refactor-nrepl/issues/333): skip scanning irrelevant directories in more places (as it was already done for various other functionalities; this limits ananlysis/refactoring to your source/test paths, skipping other artifacts).
+* Make `resolve-missing` able to find even more classes than before. 
 * [#346](https://github.com/clojure-emacs/refactor-nrepl/issues/346): refine the heuristic for ignoring irrelevant dirs (see the above bullet point).
 * Introduce `print-right-margin`/`print-miser-width` configuration options, used during `pprint`ing of ns forms.
   * The default is one that is consistent with refactor-nrepl's traditional behavior.
--- a/project.clj
+++ b/project.clj
@@ -4,6 +4,7 @@
   :license {:name "Eclipse Public License"
             :url "https://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[nrepl "0.9.0-beta3"]
+                 ^:inline-dep [compliment "0.3.12"]
                  ^:inline-dep [http-kit "2.5.3"]
                  ^:inline-dep [org.clojure/data.json "2.3.1"]
                  ^:inline-dep [org.clojure/tools.analyzer.jvm "1.1.0"]
--- a/src/refactor_nrepl/ns/class_search.clj
+++ b/src/refactor_nrepl/ns/class_search.clj
@@ -1,124 +1,17 @@
-;;;; Copied from slamhound 1.5.5
-;;;; Copyright © 2011-2012 Phil Hagelberg and contributors
-;;;; Distributed under the Eclipse Public License, the same as Clojure.
 (ns refactor-nrepl.ns.class-search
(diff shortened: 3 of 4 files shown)


bbatsov:
The only downside of doing this is that you'll double the compliment dep payload, as now compliment will get inlined twice - once in cider-nrepl and once here. As the artefacts are already quite big due to the inlining we should be careful with this. Perhaps you can use it directly from `cider-nrepl`, as `refactor-nrepl` already depends on it? Or we can assume it's one of those deps that are safe to inline as it's quite unlikely that someone will use directly compliment in their projects. Although I guess this might be an issue in dev tooling. :-)

vemv:
> The only downside of doing this is that you'll double the compliment dep payload, as now compliment will get inlined twice - once in cider-nrepl and once here.

Doesnt't seem a huge problem tbh, it means a few more KBs to be downloaded but no more much code loading time (up to two copies of `compliment.utils` will have to be `require`d. That ns has no transitive deps)

> Perhaps you can use it directly from cider-nrepl, as refactor-nrepl already depends on it?

It doen't. Both depend on Orchard though. (Both also have some duplicate deps which will get different inlining: orchard, cljfmt, tools.reader)

My suggestion would be to propose moving this code from compliment to orchard, however that requires some coordination so we can do that asynchronously.

(copy-pasting it into Orchard would seem a no-go to me - we want to reduce complexity, not increase it by having more impls, presumably maintained by different people)

bbatsov:
> It doen't.

Well, not directly, but it certainly does - https://github.com/clojure-emacs/refactor-nrepl/blob/master/project.clj#L33

I know it's not that big of a deal, but I also know how slow all the requires in the middleware are - otherwise we wouldn't have had to jump through hoops like deferred middleware loading in `cider-nrepl`. None of this was really a problem until we unlined everything. And because refactor-nrepl doesn't use the deferred loading every dep you add here makes it load even longer when it starts. Not a huge problem, but it's certainly an annoying problem. Anyways, I'm willing to merge this as it already loads so slow that most people are unlikely to notice a couple of extra seconds. 

> My suggestion would be to propose moving this code from compliment to orchard, however that requires some coordination so we can do that asynchronously.

Well, you can also inline this in `refactor-nrepl` as it was done with the old code that it replaces. Admittedly that's not a great solution, either. No need to take any action now, just wanted you to be aware that inlining more stuff is somewhat painful for the users of `refactor-nrepl`. 

For reference you can see https://github.com/clojure-emacs/cider-nrepl/issues/218 and https://github.com/clojure-emacs/cider/issues/1717

vemv:
> Well, not directly, but it certainly does - https://github.com/clojure-emacs/refactor-nrepl/blob/master/project.clj#L33

mmm. I tried now using that one's transitive `compliment`, but tests will fail (ns not found). I certainly tried with `+provided` profile. Might be a mranderson thing which I'm not quite in the mood to dive into.

> most people are unlikely to notice a couple of extra seconds.

I don't think this PR inflicts that. It bundles extra namespaces to be present in the classpath, but a namespace's _presence_ doesn't affect startup time. `(time (require 'compliment.utils :reload))` indicates it has a cost of 60ms.

> For reference you can see clojure-emacs/cider-nrepl#218 and clojure-emacs/cider#1717

Nice, I heard of those. Might be worth it to investigate whether refactor-nrepl would benefit from a similar solution? cider-nrepl + refactor-nrepl is a very common combo, so refactor-nrepl shouldn't poop anyone's party (i.e. fast-starting repl).

vemv:
Btw, refactor-nrepl is lazy I believe, I certainly remember that certain impl namespaces aren't loaded until they are used. See the `(delay (require-and-resolve ` pattern in middleware.clj, which would support that.

bbatsov:
It doesn't use the standard approach from https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl.clj#L66 `delay` is a cruder approach to the problem, that `cider-nrepl` used to some extent as well in the past. (as it's something you have to manually keep track of and it's easy to forget to do it) You should check how much extra time is added by refactor-nrepl to the server boot time, as I can't remember what the users had last reported. Anyways, updating this to a deferred wrapper will likely be easy, as if I remember correctly `refactor-nrepl` exposes just one big middleware so there's only one code change needed.

bbatsov (approved):

vemv:
Cheers, I'll remember to measure/investigate this 👍

Issue #355, opened 2021-12-11
`clj-commons/fs` changes standard clojure behaviour in `clojure.java.io/file`


ikappaki:
## Expected behavior

The REPL shouldn't change standard Clojure behaviour, otherwise it might lead to unexpected behaviour in release builds.

Brining up a REPL with clj-refactor installed allows `clojure.java.io/file` to support `java.nio.file.Path` as an argument, while it does not in the standard core, .e.g. if we bring up a vanilla REPL (wo clj-refactor) and try to open a `Path` with `io/file` it fails as expected:
```clojure
user> (-> "a-path" (java.nio.file.Paths/get (into-array String [])) clojure.java.io/file)
Execution error (IllegalArgumentException) at user/eval5725 (REPL:59).
No implementation of method: :as-file of protocol: #'clojure.java.io/Coercions found for class: sun.nio.fs.WindowsPath
```
Analysis follows at the end.

## Actual behavior

clj-refactor changes standard Clojure behaviour with regards to opening files with `clojure.java.io/file`.

## Steps to reproduce the problem

1. `M-x cider-jack-in` to a minimal `deps.edn` file.
2. At the REPL, try to open a file from a `java.nio.Path`
```clojure
user> (-> "a-path" (java.nio.file.Paths/get (into-array String [])) clojure.java.io/file)
#object[java.io.File 0x62eac79 "a-path"]
```
i.e. it returns the file from the `PATH`, while this is unsupported in standard Clojure behaviour.

### clj-refactor.el version information

clj-refactor 3.2.2 (package: 20211117.1008), refactor-nrepl 3.1.0

### CIDER version information

*Include here the version string displayed when
CIDER's REPL is launched. Here's an example:*

```
;; CIDER 1.2.0snapshot (package: 20211126.1158), nREPL 0.9.0-beta5
;; Clojure 1.10.3, Java 17
```

### Emacs version

GNU Emacs 27.2 

### Operating system

MS-Windows 10

## Analysis

The standard supported `io/file` arguments types are defined by the `io/Coercions` protocol's `as-file` method and [are `String`, `File`, `URL` and `URI`.](https://github.com/clojure/clojure/blob/abe19832c0294fec4c9c55430c9262c4b6d2f8b1/src/clj/clojure/java/io.clj#L44)

[...]

PR #356, opened 2021-12-14
Disable the extensions performed by `fs`

vemv:
These could affect end-users.

Checking out the codebase, we don't ever handle `nio.Path`s so we don't depend on them either.

Uses the flags introduced here: https://github.com/clj-commons/fs/pull/11/files

Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


expez (approved):
Nice to get this fixed 👍

vemv:
Cheers ☕️☕️

PR #360, opened 2022-01-21
Setup deployments on CI

vemv:
See also https://github.com/clojure-emacs/cider/discussions/3130

Follows a recipe that has worked well in Eastwood for example.

I also added it to https://github.com/clojure-emacs/enrich-classpath/, works (@bbatsov added the necessary CLOJARS_* variables to a few of our projects)

Updated release instructions in the readme.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,18 @@ executors:
       LEIN_ROOT: "true"   # we intended to run lein as root
       JVM_OPTS: -Xmx3200m # limit the maximum heap size to prevent out of memory errors
     <<: *defaults
+
+  # JDK 8 is preferred for deployments (https://github.com/benedekfazekas/mranderson/issues/57)
+  # Parallelism is disabled, at least for now (https://github.com/benedekfazekas/mranderson/issues/56)
+  openjdk8_deploy:
+    docker:
+      - image: circleci/clojure:openjdk-8-lein-2.9.1-node
+    environment:
+      LEIN_ROOT: "true"   # we intended to run lein as root
+      LEIN_JVM_OPTS: -Dmranderson.internal.no-parallelism=true
+      JVM_OPTS: -Xmx3200m # limit the maximum heap size to prevent out of memory errors
+    <<: *defaults
+
   openjdk11:
     docker:
       - image: circleci/clojure:openjdk-11-lein-2.9.3-buster-node
@@ -28,6 +40,7 @@ executors:
       LEIN_ROOT: "true"   # we intended to run lein as root
       JVM_OPTS: -Xmx3200m --illegal-access=deny # forbid reflective access (this flag doesn't exist for JDK8 or JDK17+)
     <<: *defaults
+
   openjdk16:
     docker:
       - image: circleci/clojure:openjdk-16-lein-2.9.5-buster-node
@@ -35,6 +48,7 @@ executors:
       LEIN_ROOT: "true"   # we intended to run lein as root
       JVM_OPTS: -Xmx3200m --illegal-access=deny # forbid reflective access (this flag doesn't exist for JDK8 or JDK17+)
     <<: *defaults
+
   openjdk17:
     docker:
       - image: circleci/clojure:openjdk-17-lein-2.9.5-buster-node
@@ -112,6 +126,24 @@ jobs:
           cache_version: "1.10"
           steps: << parameters.steps >>
 
+  deploy:
+    executor: openjdk8_deploy
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v2-dependencies-{{ checksum "project.clj" }}
+            # fallback to using the latest cache if no exact match is found
+            - v2-dependencies-
+      - run: lein with-profile -user,+test deps
+      - save_cache:
+          paths:
+            - ~/.m2
+          key: v2-dependencies-{{ checksum "project.clj" }}
+      - run:
+         name: Deploy
+         command: |
+           lein with-profile -user,+deploy run -m maybe-deploy make deploy
 
   test_code:
     description: |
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -21,20 +21,34 @@ executors:
...
     <<: *defaults
+
+  # JDK 8 is preferred for deployments (https://github.com/benedekfazekas/mranderson/issues/57)
+  # Parallelism is disabled, at least for now (https://github.com/benedekfazekas/mranderson/issues/56)
+  openjdk8_deploy:
+    docker:
+      - image: circleci/clojure:openjdk-8-lein-2.9.1-node
+    environment:
+      LEIN_ROOT: "true"   # we intended to run lein as root
+      LEIN_JVM_OPTS: -Dmranderson.internal.no-parallelism=true
```
vemv, line 32: Note that parallelism is disabled only in deployments, and because they're performed on CI, the slowness don't matters as much - one doesn't have to stare at the output 😄

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 3.2.1 (2022-01-21)
```
vemv, line 5: Included sort of a dummy change so that we can try the new setup (which can be done once we merge this PR; before tags will be ignored since they are in `master`)

vemv:
Added some last tweaks, PR ready

vemv:
Added some tweaks from what I learned today deploying https://github.com/clojure-emacs/enrich-classpath with an analog setup.

Ready again

.circleci/config.yml
bbatsov: you might want to add a comment about this in the config.

.circleci/config.yml
```diff
@@ -21,20 +21,34 @@ executors:
       LEIN_ROOT: "true"   # we intended to run lein as root
       JVM_OPTS: -Xmx3200m # limit the maximum heap size to prevent out of memory errors
     <<: *defaults
+
+  # JDK 8 is preferred for deployments (https://github.com/benedekfazekas/mranderson/issues/57)
+  # Parallelism is disabled, at least for now (https://github.com/benedekfazekas/mranderson/issues/56)
+  openjdk8_deploy:
+    docker:
+      - image: circleci/clojure:openjdk-8-lein-2.9.1-node
```
bbatsov, line 29: Are there no images with 2.9.3? Not a big deal, just looks a bit weird to use a different version of Lein for different JDKs.

bbatsov: Actually it seems that 2.9.5 is the newest.

.circleci/maybe_deploy.clj
```diff
@@ -0,0 +1,32 @@
+(ns maybe-deploy
```
bbatsov, line 1: maybe-deploy? Sound a bit confusing to me. Perhaps `deploy-release` or something like this?

bbatsov:
I probably wouldn't remove the manual deployment stuff. I'd just advise against their usage if the CI deployment is working as it should be.

vemv:
> I probably wouldn't remove the manual deployment stuff. I'd just advise against their usage if the CI deployment is working as it should be

IDK, using CI ensures various important stuff (using the right jdk, no parallelism, verify that the tests are passing) that can be easily forgotten, say, 1y from now in a rush.

Probably as long as we keep cutting releases at a relatively slow pace, it doesn't matter a lot if a release cannot, hypothetically, be _immediately_ deployed (b/c a red build or issue with the setup).

In practice I've been very happy with the analog setup that Eastwood has - would never look back.

Finally, a savvy developer can always hack-deploy things even if undocumented. But I'd really try to go in the opposite direction: strongly discourage hack-deploys because we want to publish dependable software for a variety of consumers.

.circleci/config.yml
vemv: I think that with `-node` it only goes up to 2.9.1, for this jdk version

And probably `-node` is needed, as it certainly is for the test suite and maybe the mrandersoning?

.circleci/config.yml
vemv: There are some comments in L25

.circleci/maybe_deploy.clj
vemv: Renamed

bbatsov:
Fair enough. I'm mostly concerned that historically the CI was very unstable (due to hard to reproduce issues), but perhaps that's behind us now. More motivation to make sure that it's always green. :-)

resources/refactor_nrepl/version.edn
```diff
@@ -0,0 +1,3 @@
+;; This file is automatically overwriten from .circleci/maybe_deploy.clj,
```
bbatsov, line 1: You should update the `maybe_deploy` reference here.

project.clj
```diff
@@ -1,16 +1,17 @@
-(defproject refactor-nrepl "3.2.0"
+(defproject refactor-nrepl (or (not-empty (System/getenv "PROJECT_VERSION"))
```
bbatsov, line 1: What's the point of the fallback version here? I get why the code is like this, it just look a bit confusing. Might be good to mention that the version will be set by `deploy-release` and just use `nil` as the placeholder in case PROJECT_VERSION is not set.

.circleci/config.yml
bbatsov: I see. I guess down the road we can just install whatever else is needed ourselves in a more stripped down container, so that the setup in more consistent.

bbatsov (approved):
Apart from my small remarks everything looks good.

project.clj
vemv: Probably `nil` can cause problems wherever a string was expected, might have been the case from what I vaguely remember

Issue #359, opened 2022-01-21
Dont remove ns requires without an alias


dpsutton:
Clj-kondo and refactor-nrepl disagree about how to clean `ns` forms and it would be nice if we could unify their notions. The problem is around how to tell the linter to not remove a namespace require which has no visible usages in the buffer, but is required for side effects. Refactor nrepl will remove them unconditionally and there seems to be an agreed upon workaround to include a form

```clojure
(ns (:require [foo.bar :as bar]))
(comment
  (bar/keep-me))
```

Clj-kondo takes the approach that any require that does not have an alias is presumably there for side effects.

```clojure
(ns foo
  (:require [foo.bar]
            foo.bar.baz))
```

In this form both `foo.bar` and `foo.bar.baz` will remain because they do not have aliases.

This analysis also powers how `clojure-lsp` will clean the ns form.

It would be nice if refactor-nrepl would align to this style. Clj-kondo is by far the most popular linter and will complain about the unresolved symbols in the `bar/keep-me` form. Clojure-lsp is powered by clj-kondo's analysis and will count those alias use sites as usages and keep the requires they intend to keep, but will also keep them if the alias is dropped since it uses clj-kondo's analysis. Adopting this style in refactor-nrepl would mean less work for users of refactor-nrepl and no tension with the other tooling.

borkdude:
Note that there are two conditions: no alias or no referred vars.

magnars:
Agreed. I remember this being discussed here before, but I don't remember why it wasn't changed. Maybe @expez has a clue?

borkdude:
Found an issue here: https://github.com/clojure-emacs/refactor-nrepl/issues/292

expez:
> Agreed. I remember this being discussed here before, but I don't remember why it wasn't changed. Maybe @expez has a clue?

Nobody's added the setting.  That's the only reason 🙂   Personally I prefer the workaround in #292 (and I don't write much cljs!) because I think it's helpful to indicate that the required ns is special and I also always add a comment about why we need those side-effects.

vemv:
From https://github.com/clj-kondo/clj-kondo/issues/241 which dates back to 2019 a few things can be observed:

* The most-voted option was ignored
* The chosen option was acknowledged to yield false negatives
* These days, projects can have kondo config that would be propagated to lib consumers
  * This can be used to mark side-effectful namespaces, which could be part of clj-kondo's logic

[...]

PR #361, opened 2022-01-29
Honor clj-kondo `:unused-namespace` config, if present

vemv:
> Context: https://github.com/clojure-emacs/refactor-nrepl/issues/359#issuecomment-1019639865

Proposes to read .clj-kondo config in search of `:unused-namespace` config which is essentially the same as our `:libspec-whitelist` config option.

This way, refactor-nrepl can discover existing config, keeping things smart/DRY, and also reducing friction for teams using diverse tooling.

Feedback welcome - not necessarily to merge as-is!

### DIFF
--- a/src/refactor_nrepl/ns/libspec_whitelist.clj
+++ b/src/refactor_nrepl/ns/libspec_whitelist.clj
@@ -15,7 +15,9 @@
                         (throw e)))))]
     (->> exclude
          (mapv (fn [entry]
-                 (re-pattern (str "^" (Pattern/quote (str entry)) "$"))))
+                 (if (symbol? entry)
+                   (str "^" (Pattern/quote (str entry)) "$")
+                   entry)))
          (into (:libspec-whitelist config/*config*)))))
 
 (def ^:private ^:dynamic *libspec-whitelist* nil)
--- a/src/refactor_nrepl/ns/libspec_whitelist.clj
+++ b/src/refactor_nrepl/ns/libspec_whitelist.clj
@@ -1,0 +1,47 @@
+(ns refactor-nrepl.ns.libspec-whitelist
+  (:require
+   [clojure.java.io :as io]
+   [refactor-nrepl.config :as config])
+  (:import
+   (java.util.regex Pattern)))
+
+(defn- libspec-whitelist* []
+  (let [kondo-file (io/file ".clj-kondo" "config.edn")
+        exclude (when (.exists kondo-file)
+                  (try
+                    (-> kondo-file slurp read-string :linters :unresolved-namespace :exclude)
+                    (catch Exception e
+                      (when (System/getenv "CI")
+                        (throw e)))))]
+    (->> exclude
+         (mapv (fn [entry]
+                 (re-pattern (str "^" (Pattern/quote (str entry)) "$"))))
+         (into (:libspec-whitelist config/*config*)))))
+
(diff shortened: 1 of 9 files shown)


src/refactor_nrepl/ns/libspec_whitelist.clj
```diff
@@ -0,0 +1,47 @@
...
+  (let [kondo-file (io/file ".clj-kondo" "config.edn")
+        exclude (when (.exists kondo-file)
+                  (try
+                    (-> kondo-file slurp read-string :linters :unused-namespace :exclude)
+                    (catch Exception e
+                      (when (System/getenv "CI")
+                        (throw e)))))]
+    (->> exclude
+         (mapv (fn [entry]
+                 (re-pattern (str "^" (Pattern/quote (str entry)) "$"))))
```
vemv, line 18: I shouldn't Pattern/quote this if it's already a regex-like string. Official example:

```clj
{:linters {:unused-namespace {:exclude [".*\\.specs$"]}}}
```

bbatsov:
I'm fine with the proposal. Just make sure it's documented in the docs as well, so people won't be surprised.

src/refactor_nrepl/ns/libspec_whitelist.clj
```diff
@@ -0,0 +1,49 @@
+(ns refactor-nrepl.ns.libspec-whitelist
```
bbatsov, line 1: Let's avoid the problematic "whitelist" terminology. I guess the namespace can be simply named "libspec".

vemv: The nice thing about the name is that it maps 1:1 to the existing piece of config.

FWIW I also favor avoiding these terminologies where I can, but often keeping things maintainable/non-broken takes precedence.

bbatsov: I know how the config value is named, but I don't see any reason to name a namespace after a configuration option. Especially after in the long run it's highly likely that the current name will be deprecated in favor of some alias with a different name.

magnars:
If some alternative to whitelist is useful, let me suggest allowlist.

bbatsov:
Yeah, that's the most common alternative. Generally I prefer the more specific "allowed-libspecs" or "ignored-libspecs". There's also the part that the name doesn't really imply it's related to `clean-ns`, so a name like `clean-ns-allowed-libspecs` would be even better IMO.

vemv:
> , but I don't see any reason to name a namespace after a configuration option.

It would mean a less ample vocabulary one has to learn and fewer mappings to keep in one's head.

Anyway I've renamed the internals to `*allowlist`. The piece of config itself can be renamed later.

bbatsov:
Btw, I still don't get why we need a separate ns for this, as there's already a `libspecs` ns where it seems to me that this would fit. I've never been too fond of overly granular namespaces, unless there are needed to avoid some circular deps.

vemv:
Yeah that's one of those things that we'll never quite agree on :)

For me an objective decider is that tools.namespace `(refresh)` works better with fine-grained namespaces, otherwise altering one file can cascade into reloading many more namespaces, which is slower.

bbatsov:
I get your point, but I've always aimed to optimize code for people not for tools. :-) Anyways, not a big deal.

PR #367, opened 2022-02-08
Don't prune `require` forms if they are needed for a given `import` to work

vemv:
Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/194

### DIFF
--- a/src/refactor_nrepl/ns/prune_dependencies.clj
+++ b/src/refactor_nrepl/ns/prune_dependencies.clj
@@ -1,6 +1,8 @@
 (ns refactor-nrepl.ns.prune-dependencies
   (:require
    [cider.nrepl.middleware.info :as info]
+   [clojure.set :as set]
+   [clojure.string :as string]
    [refactor-nrepl.core :as core]
    [refactor-nrepl.find.symbols-in-file :as symbols-in-file]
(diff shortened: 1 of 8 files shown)


src/refactor_nrepl/ns/prune_dependencies.clj
```diff
@@ -116,18 +118,69 @@
                      (-> pattern re-pattern (re-find ns-name)))
                    (libspec-allowlist/libspec-allowlist)))))
 
-(defn- prune-libspec [symbols-in-file current-ns libspec]
-  (if (libspec-should-never-be-pruned? libspec)
+(defn imports->namespaces [imports]
```
bbatsov, line 121: All public functions should have a docstring (and ideally an `:added` metadata as well).

src/refactor_nrepl/ns/prune_dependencies.clj
```diff
@@ -116,18 +118,69 @@
...
+                     (first import)
+                     (->> (-> import str (string/split #"\."))
+                          (butlast)
+                          (string/join ".")))
+                   str
+                   (string/replace "_" "-")
+                   symbol)))
+        imports))
+
+(defn libspec->namespaces [libspec]
```
bbatsov, line 134: Same here.

bbatsov (approved):
Overall the changes look good to me.

Issue #142, opened 2016-02-09
File not found when using move form refactoring


scramjet:
Hello, I'm trying to move a function to a new namespace ("move one or more forms to another namespace, :refer any functions"), but getting the exception below (note that this is from Emacs' `*Messages*` buffer and includes the fact that the file it's complaining about actually exists).

The function is moved and the target namespace has some new require's added, but nothing else referring to it is updated.

I'm using clj-refactor 2.0 and the latest stable CIDER.

Any ideas, or things I should try?

```
Wrote /Users/matt/Development/basis/src/basis/network/core.clj
cljr--get-error-value: Error in nrepl-refactor: java.io.FileNotFoundException: src/basis/network/core.clj (No such file or directory)
 at java.io.FileInputStream.open0 (FileInputStream.java:-2)
    java.io.FileInputStream.open (FileInputStream.java:195)
    java.io.FileInputStream.<init> (FileInputStream.java:138)
    java.io.FileInputStream.<init> (FileInputStream.java:93)
    java.io.FileReader.<init> (FileReader.java:58)
    sun.reflect.NativeConstructorAccessorImpl.newInstance0 (NativeConstructorAccessorImpl.java:-2)
    sun.reflect.NativeConstructorAccessorImpl.newInstance (NativeConstructorAccessorImpl.java:62)
    sun.reflect.DelegatingConstructorAccessorImpl.newInstance (DelegatingConstructorAccessorImpl.java:45)
    java.lang.reflect.Constructor.newInstance (Constructor.java:422)
    clojure.lang.Reflector.invokeConstructor (Reflector.java:180)
    refactor_nrepl.core$read_ns_form.invokeStatic (core.clj:234)
    refactor_nrepl.core$read_ns_form.invoke (core.clj:234)
    refactor_nrepl.ns.clean_ns$clean_ns.invokeStatic (clean_ns.clj:46)
    refactor_nrepl.ns.clean_ns$clean_ns.invoke (clean_ns.clj:38)
    refactor_nrepl.middleware$clean_ns_reply.invokeStatic (middleware.clj:75)
    refactor_nrepl.middleware$clean_ns_reply.invoke (middleware.clj:74)
    refactor_nrepl.middleware$wrap_refactor$fn__18008.invoke (middleware.clj:123)
    clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__721.invoke (middleware.clj:22)
[...]

PR #368, opened 2022-02-09
`read-ns-form`: rethrow `FileNotFoundException`s more informatively

vemv:
* `read-ns-form`: rethrow `FileNotFoundException`s more informatively
  * Simply rethrows this exception if it arises with particularly useful wording and ex-data. Otherwise logic isn't changed at all.
  * Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/142
* Bump Eastwood


src/refactor_nrepl/core.clj
```diff
@@ -152,28 +152,28 @@
 
 (defn read-ns-form
   ([path]
-   (let [^String path-string (when (string? path)
```
vemv, line 155: DRYed out the duplication across the two arities of `read-ns-form`.

Took care of making them 100% equivalent. The key is here:

```clj
(if dialect
  {:read-cond :allow :features #{dialect}}
  nil)
```

The nil value makes `readers/indexing-push-back-reader` behave exactly as if it hadn't received this argument at all.

Issue #369, opened 2022-02-15
Make `namespace-aliases` aware of new files


vemv:
## Context

https://stuartsierra.com/2015/05/10/clojure-namespace-aliases is a relatively popular pattern of aliasing.

For a ns named `foo.bar.baz`, `baz` or `bar.baz` would be idiomatic aliases.

For a _newly created_ namespace `foo.bar.baz`, no other namespace will have aliased it yet, because it's new.

Yet, as a clj-refactor.el user, if I type e.g. `baz <slash>` i.e. `cljr-slash`, ideally I would get completions for inserting a `foo.bar.baz` :require libspec.

## Task

Make [namespace-aliases](https://github.com/clojure-emacs/refactor-nrepl/tree/6529ec03c854eddcd0b62f359378d066e9d88d3d#namespace-aliases) also return tentative Sierra-style aliases if:

* they are not taken already by an existing alias
* there exist files that would satisfy this alias (e.g. `foo/bar/baz.clj` exists)
  *  _and_ this file is not aliased in any dependent namespaces.
* an opt-in option was passed.

n.b. a newly created ns may not be `load`ed yet.

PR #370, opened 2022-02-21
Implement "suggest" option for the `namespace-aliases` op

vemv:
Fixes https://github.com/clojure-emacs/refactor-nrepl/issues/369

I've tried it over my local clj-refactor.el copy and it works smoothly and quickly, aided by the usual measures we take for performance (parallelism, caching)

The gist of the feature is that if I create a _new_ namespace called `my.prefix.foo`, then in another ns I'll be able to type `foo <slash>`, and cljr-slash will insert a `my.prefix.foo :as foo` libspec.

`prefix.foo` would also work, i.e. both `foo` and `prefix.foo` are suggested aliases for `my.prefix.foo`.

Other minor tweaks also included in the PR, one per commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* [#369](https://github.com/clojure-emacs/refactor-nrepl/issues/369): Implement "suggest" option for the `namespace-aliases` op.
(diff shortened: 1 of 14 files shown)


vemv:
This one has worked well for me over the last few days.

PR can be reviewed

PR #372, opened 2022-03-08
Extract `namespace-aliases-for`

vemv:
As indicated in an inline comment:

`namespace-aliases-for` was split out from `namespace-aliases`, for a 3rd-party need.

`namespace-aliases-for` is a little more fine-grained, since it accepts files rather than dirs.

### DIFF
--- a/src/refactor_nrepl/ns/libspecs.clj
+++ b/src/refactor_nrepl/ns/libspecs.clj
@@ -95,6 +95,33 @@
          (apply dissoc possible-aliases)
          (merge-with into project-aliases))))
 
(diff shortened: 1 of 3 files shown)


src/refactor_nrepl/ns/libspecs.clj
```diff
@@ -110,29 +137,7 @@
   ([ignore-errors? dirs include-tentative-aliases?]
    (let [;; fetch the file list just once (as opposed to traversing the project once for each dialect)
          files (core/source-files-with-clj-like-extension ignore-errors? dirs)
-         ;; pmap parallelizes a couple things:
```
vemv, line 113: This code has been extracted as-is to the new defn, without any creative change.

Issue #384, opened 2022-08-20
Implement new middleware op: `suggest-libspecs`


vemv:
https://github.com/clojure-emacs/refactor-nrepl/tree/v3.5.4#namespace-aliases is currently used for cljr-slash. As the op name says, it returns the namespaces aliases contained in the current project.

However, a more sophisticated (and context-aware, for .cljc files) `cljr-slash` needs more than simply a project map. We want to compute _suggestions_, based on logic, besides from project data.

The best place to perform this logic is refactor-nrepl (vs. clj-refactor), so we'd need a new piece of middleware. This prevents `namespace-aliases` from being bloated and possibly inflicting breaking changes.

As an additional benefit, the middleware responses for cljr-slash will be much smaller (just the useful info vs. a huge map, 99% of which will be unused).

## The `suggest-libspecs` middleware op

suggest-libspecs takes as inputs:

* a lib prefix (e.g. `io`, representing what the user just typed `io/`, triggering cljr-slash)
* a language context
  * one of: clj, cljs, cljc
  * Represents the extension of the user's filename
* A set of preferred aliases
  * Represents the whole value of the `cljr-magic-require-namespaces` defcustom
  * See also: https://github.com/clojure-emacs/clj-refactor.el/issues/530

out of those inputs, and of project analysis, it decides which libspec(s) can be suggested to the user.

The output format is as follows:

```clj
["[namespace :as alias]"
 "[another-namespace :as alias]"]
```

i.e. it is a simple sequence of choices, ordered by inferred likelihood of choice.

There is no need for metadata e.g. `{:context :clj}`, since the middleware already returns choices that are valid for the provided inputs.

choices might be expressed as reader conditionals:

```clj
["[#?(:clj foo.clj :cljs foo.cljs) :as alias]"]
```

...such reader conditionals would reflect the already-used reader conditionals from a given project, or be synthetically built when it makes sense.

.choices can include `:refer` clauses in a couple cases:

* for clojure.test in particular, where often one always wants the same :refer clause
* when it would enable a valid/useful reader conditional

other than that, a generalized inclusion of `:refer`s is not foreseen.

...and similarly, `:include-macros true` clauses can be included.

When nothing can be suggested, an empty sequence `[]` will be returned.

Suggestions are suject to caching.

#### 1

```
Given the user intent `set/`:
When [clojure.set :as set] is specified in clj-refactor's cljr-magic-require-namespaces as OK in both contexts: clj, cljs
[...]

PR #385, opened 2022-08-20
Don't suggest alias fragments with leading number

pdbrown:
Hey there! By pure coincidence, I happened to be investigating a possible bug in `cljr-slash` and `refactor-nrepl.ns.suggest-aliases/suggested-aliases` just as you posted this writeup #384 today, which looks great btw!
I have a project with some namespaces where the last segment starts with a number like `db.migration.005-alter-table`. `suggested-aliases` offers `005-alter-table` as a possible alias, but that's not a valid suggestion since the clojure reader doesn't accept symbols that start with a number. When the nrepl respose is parsed by parseclj, it also correctly rejects that symbol, but that breaks cljr-slash for the entire project.
I put together this little change to remove such invalid symbols from the suggestions, and it's working well for me so far!

Thanks!

### DIFF
--- a/src/refactor_nrepl/ns/suggest_aliases.clj
+++ b/src/refactor_nrepl/ns/suggest_aliases.clj
@@ -1,6 +1,7 @@
 (ns refactor-nrepl.ns.suggest-aliases
   "Suggestion of aliases based on these guidelines: https://stuartsierra.com/2015/05/10/clojure-namespace-aliases"
   (:require
+   [clojure.edn :as edn]
    [clojure.string :as string]))
 
 (defn test-like-ns-name? [ns-sym]
@@ -11,6 +12,14 @@
                    (or (string/starts-with? last-fragment "t-")
                        (string/starts-with? last-fragment "test-")
                        (some #{"test" "unit" "integration" "acceptance" "functional" "generative"} fragments)))))))
+
+(def readable-as-symbol?
+  (memoize
+   (fn [s]
+     (try
+       (symbol? (edn/read-string s))
+       (catch Exception _
+         false)))))
 
 (defn suggested-aliases [namespace-name]
   (let [fragments (-> namespace-name str (string/split #"\."))
@@ -29,8 +38,10 @@
                        (range 1 (inc (count fragments)))
                        (repeat (distinct fragments)))
         v (into {}
-                (map (fn [segments]
-                       [(->> segments (string/join ".") (symbol)),
-                        [namespace-name]]))
+                (keep (fn [segments]
+                        (let [candidate (->> segments (string/join "."))]
+                          (when (readable-as-symbol? candidate)
+                            [(symbol candidate),
+                             [namespace-name]]))))
                 fragments)]
     (dissoc v namespace-name)))
--- a/test/refactor_nrepl/ns/suggest_aliases_test.clj
+++ b/test/refactor_nrepl/ns/suggest_aliases_test.clj
@@ -21,6 +21,25 @@
     'unit.foo   true
     'foo.unit   true))
 
+(deftest readable-as-symbol?
+  (testing "is readable as symbol"
+    (are [input] (sut/readable-as-symbol? input)
+      "test"
+      "a.test"
+      "a-test"
+      "b.a-test"
+      "ok.5<-here"
+      "this:too!#$%&*"))
+
+  (testing "not readable as symbol"
+    (are [input] (not (sut/readable-as-symbol? input))
+      ":kw"
+      "15"
+      "5-bad"
+      "6:nope"
+      ";what"
+      "#yeah")))
+
 (deftest suggested-aliases
   (are [desc input expected] (testing input
                                (is (= expected
@@ -46,4 +65,7 @@
     "Removes redundant bits such as `clj-` and `.core`"
     'clj-a.b.c.core '{c     [clj-a.b.c.core]
                       b.c   [clj-a.b.c.core]
-                      a.b.c [clj-a.b.c.core]}))
+                      a.b.c [clj-a.b.c.core]}
+
+    "Removes fragments that start with a number (in this case: `9d`, `8b.c.9d`), since they aren't valid symbols"
+    'a.8b.c.9d '{c.9d [a.8b.c.9d]}))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* [#385](https://github.com/clojure-emacs/refactor-nrepl/pull/385): Fix cljr-refactor.el's `cljr-slash` when namespace fragments start with a number.
+
 ## 3.5.4
 
 * [#383](https://github.com/clojure-emacs/refactor-nrepl/issues/383): `prune-dependencies`: increase accuracy for `:cljs`.


test/refactor_nrepl/ns/suggest_aliases_test.clj
```diff
@@ -46,4 +46,7 @@
     "Removes redundant bits such as `clj-` and `.core`"
     'clj-a.b.c.core '{c     [clj-a.b.c.core]
                       b.c   [clj-a.b.c.core]
-                      a.b.c [clj-a.b.c.core]}))
+                      a.b.c [clj-a.b.c.core]}
+
+    "Removes fragments that start with a number, since they aren't valid symbols"
```
vemv, line 51:
```suggestion
    "Removes fragments that start with a number (in this case: `9d`), since they aren't valid symbols"
```

vemv (approved):
Thanks much!

Kudos for finding the accurate cause of your problem.

Will be releasing this today as we merge this.

src/refactor_nrepl/ns/suggest_aliases.clj
```diff
@@ -28,6 +28,9 @@
         fragments (map take-last
                        (range 1 (inc (count fragments)))
                        (repeat (distinct fragments)))
+        fragments (remove (fn [segments]
+                            (re-find #"^\d+" (first segments)))
+                          fragments)
```
vemv, line 33:
```suggestion
```

Let's do the `remove` a few lines below:

```clj
(into {}
      (keep (fn [segments]
              (let [candidate (->> segments (string/join "."))]
                (when-not (invalid-symbol? candidate)
                  [(symbol candidate),
                   [namespace-name]]))))
      fragments)
```

where: 

```clj
(def invalid-symbol? (memoize (fn [x]
                                (try
                                  (edn/read-string x)
                                  false
                                  (catch Exception _
                                    true)))))
```

This way we gain genericity (my no. 1 concern - I don't want people to encounter similar problems in the future) and possibly performance.

Please make sure to unit-test `invalid-symbol?`

Thanks!

vemv: (feel free to find a better name for `invalid-symbol?` - it accepts strings not symbols)

pdbrown: Sure thing! I added a `symbol?` test too, since we probably shouldn't accept keywords etc either.

test/refactor_nrepl/ns/suggest_aliases_test.clj
```diff
@@ -4,8 +4,8 @@
    [refactor-nrepl.ns.suggest-aliases :as sut]))
 
 (deftest test-like-ns-name?
-  (are [input expected] (= expected
-                           (sut/test-like-ns-name? input))
+  (are [input expected] (is (= expected
```
vemv, line 7: the `is` is implied, please remove

test/refactor_nrepl/ns/suggest_aliases_test.clj
```diff
@@ -21,6 +21,22 @@
     'unit.foo   true
     'foo.unit   true))
 
+(deftest readable-as-symbol?
+  (are [input expected] (is (= expected
```
vemv, line 25: no `is`

test/refactor_nrepl/ns/suggest_aliases_test.clj
```diff
@@ -46,4 +62,7 @@
     "Removes redundant bits such as `clj-` and `.core`"
     'clj-a.b.c.core '{c     [clj-a.b.c.core]
                       b.c   [clj-a.b.c.core]
-                      a.b.c [clj-a.b.c.core]}))
+                      a.b.c [clj-a.b.c.core]}
+
+    "Removes fragments that start with a number (in this case: `9d`), since they aren't valid symbols"
```
vemv, line 67:
```suggestion
    "Removes fragments that start with a number (in this case: `9d`, `8b.c.9d`), since they aren't valid symbols"
```

test/refactor_nrepl/ns/suggest_aliases_test.clj
```diff
@@ -21,6 +21,22 @@
...
     'foo.unit   true))
 
+(deftest readable-as-symbol?
+  (are [input expected] (is (= expected
+                               (sut/readable-as-symbol? input)))
+    "test"     true
+    "a.test"   true
+    "a-test"   true
+    "b.a-test" true
+    "ok then"  true
```
vemv, line 31: while I don't expect `ok then` be produced irl, it is a good edge case to think of.

If it's at hand, the result should be `false` because a symbol with a space can be surprising / unusable client-side.

Then again I don't expect it to be a real problem, so feel free not to spend time in it

(in which case, remove this line so that we remain agnostic)

vemv (commented):
> Sure thing! I added a symbol? test too, since we probably shouldn't accept keywords etc either.

Good thinking, thanks!

test/refactor_nrepl/ns/suggest_aliases_test.clj
pdbrown: Oh yes, that was a lazy hack to see the failing line `expected: (is (= false (sut/readable-as-symbol? "ok.one.more!")))` vs `expected: false`. I put in a proper fix: a plain `(readable-as-symbol? expected)` does the trick vs `(is (= true (...)))`.

test/refactor_nrepl/ns/suggest_aliases_test.clj
pdbrown: Good point, took it out.

test/refactor_nrepl/ns/suggest_aliases_test.clj
vemv: Yep, I understand!

That's why there's a `true` at tail position in this ns's `are`s (kondo and Eastwood won't complain - they understand the trick)

vemv:
Thanks much and kudos for driving it to completion!

I'll release these today and you can count on it being available by tomorrow morning.

btw, I don't know if you were using the [suggest](https://github.com/clojure-emacs/clj-refactor.el/blob/1bd7b09976c05ebd1ff1ee3d5c882af303780b67/clj-refactor.el#L1945)  feature actively. I am, and it's been such a nice QoL improvement.

https://github.com/clojure-emacs/refactor-nrepl/issues/384 will be interesting as well, backing a revamped `cljr-slash`.

vemv:
Available in clj-refactor (3.5.6; will be released within a couple hours) / refactor-nrepl (3.5.5; released)

pdbrown:
Awesome, and thanks for all the quick reviews! Yeah, I just started using it and it's very helpful (which is how I noticed it wasn't working in my biggest project). I actually recently enabled clj-refactor again after not using it for a few years, and it's so great! I'd disabled it back in the day when I couldn't figure out how to prevent it from building ASTs, and perf was killing me due to my slow laptop + frequent repl restarts before I'd learned how to make things more reloadable.
The new suggest-libspecs op looks cool too! I'm interested in CLJC awareness and also smaller nREPL responses. I've noticed CIDER's `bdecode` can create a lot of garbage, so anything we do to shrink responses should help.

vemv:
Awesome! That's great and motivating input to hear.

AST building can indeed be slow, it cannot scale indefinitely but that's just how `tools.analyzer` is.

Normally I work on 'modular' projects anyway so no codebase will be overly big.

We also possibly have a protocol that would use kondo / lsp instead. There's a lot to be gained with such hybrids.

Issue #292, opened 2021-02-08
clean-ns should follow the convention to not prune namespace deps when they don't have a :refer or :as definition


RickMoynihan:
The convention came about in this clj-kondo thread:

https://github.com/clj-kondo/clj-kondo/issues/241

Essentially the conclusion is to protect namespace declarations that are side-effecting from pruning (e.g. namespaces that load protocol extensions or multimethods etc) by not having them `:refer` anything or use an `:as` form.

clj-kondo and I believe some other tools already follow this convetion; it would be nice for `refactor-nrepl` to support this too.

I suspect it would be pretty simple to edit this function to support it?

https://github.com/clojure-emacs/refactor-nrepl/blob/dbafd6e686402dc020d7026e21ee556812fc51da/src/refactor_nrepl/ns/prune_dependencies.clj#L118-L122

bbatsov:
PR welcome!

expez:
We've discussed this already in this repo and I consider it a non-issue for Clojure code.  As I said over [here](https://github.com/clojure-emacs/refactor-nrepl/pull/270) I solve this by doing:

```clj
(ns my-ns
  (:require [clojure.string :as str]))

;; This has some really sweet side-effects that we want, because reasons
(require 'tick.timezone) 
```

This makes it super clear that this ns needs special consideration and requires no out-of-band knowledge about conventions.

Since people keep asking for this feature (and the solution above doesn't work for cljs) I agree with @bbatsov that a PR would be great, with the caveat that it has to be opt-in via a new setting.  That way we don't create confusion for our existing user-base and leave the door open for the solution above that I find preferable when dealing with a Clojure-only project.

vemv:
PR certainly welcome (with the mentioned constraints: should be opt-in)

Closing the issue as some time has passed, it's more manageable to have few issues in the tracker.

PR #386, opened 2022-09-17
Add setting that allows keeping unused unaliased requires

OknoLombarda:
> https://github.com/clojure-emacs/refactor-nrepl/issues/292
> https://github.com/clojure-emacs/refactor-nrepl/issues/359

Add setting that allows keeping unused unaliased requires, disabled by default

- [ ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)

Thanks!

### DIFF
--- a/src/refactor_nrepl/ns/prune_dependencies.clj
+++ b/src/refactor_nrepl/ns/prune_dependencies.clj
@@ -6,7 +6,8 @@
    [refactor-nrepl.core :as core]
    [refactor-nrepl.find.symbols-in-file :as symbols-in-file]
    [refactor-nrepl.ns.libspec-allowlist :as libspec-allowlist]
-   [refactor-nrepl.util :as util]))
+   [refactor-nrepl.util :as util]
+   [refactor-nrepl.config :as config]))
 
 (defn- lookup-symbol-ns
   ([ns symbol]
@@ -60,7 +61,9 @@
                       symbols-in-file)
                 (some (partial libspec-in-use-without-refer-all? libspec)
                       symbols-in-file))
-              (libspec-in-use-with-rename? libspec symbols-in-file))
+              (libspec-in-use-with-rename? libspec symbols-in-file)
+              (and (not libspec-as)
+                   (:keep-unused-unaliased-requires config/*config*)))
       libspec)))
 
 (defn- referred-symbol-in-use?
--- a/src/refactor_nrepl/config.clj
+++ b/src/refactor_nrepl/config.clj
@@ -21,6 +21,9 @@
    ;; sort etc
    :prune-ns-form true
 
+   ;; When true `clean-ns` will not remove unused requires without aliases
+   :keep-unused-unaliased-requires false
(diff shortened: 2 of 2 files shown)


OknoLombarda:
I'll add tests and update readme if you have no comments regarding the code

src/refactor_nrepl/ns/prune_dependencies.clj
```diff
@@ -60,7 +61,9 @@
                       symbols-in-file)
                 (some (partial libspec-in-use-without-refer-all? libspec)
                       symbols-in-file))
-              (libspec-in-use-with-rename? libspec symbols-in-file))
+              (libspec-in-use-with-rename? libspec symbols-in-file)
+              (and (not libspec-as)
```
expez, line 65: The example in [clj-kondo-linters](https://github.com/clj-kondo/clj-kondo/blob/master/doc/linters.md#unused-namespace) use `(ns foo (:require [foo.specs]))`.

In other words, this will allow both `foo.specs` and `[foo.specs]` to remain, but isn't the wrapping what's used to signal that it's side-effecting in the clj-kondo convention?

OknoLombarda: It doesn't seem like clj-kondo cares about brackets 
image

vemv:
The resolution of https://github.com/clojure-emacs/refactor-nrepl/issues/359#issuecomment-1032401701 is quite more recent than the one of the other issue.

I'd like refactor-nrepl to contribute to a more explicit, standarized tooling landscape.

The lack of an alias does't mean much in official Clojure terms. The extra semantic is a clj-kondo invention, that precedes its more explicit `:unused-namespace` config.

So I'd prefer that people write explicit .clj-kondo config. It's something that refactor-nrepl currently reads. It also is something that other tools (e.g. Eastwood) could also eventually read.

It's a much better approach than reading magic comments (`#_,,,`) or making up semantics. For users and tool makers alike.

vemv:
Also, non trivially, one approach is naturally more prone to false positives than the other.

If I add an entry to my .clj-kondo file, then my intent is certain.

Contrariwise, adding `[some-lib]` without an alias is an everyday action, it can happen for any reason, from any person. So we can easily get a false positive.

For that reason alone, e.g. making our tools _correct_ I think the right approach is clear now.

Will be moving this PR to draft (with the intent of closing it if there isn't much opposition)

OknoLombarda:
Yeah, I just though that'd be convenient, but you're right, picking convenience in some situations over correctness is not worth it `【⸟ ͟ʖ⸟】`

magnars:
Any reason this could not be extended to actually using the clj-kondo file? Or is that already supported?

OknoLombarda:
It is already supported https://github.com/clojure-emacs/refactor-nrepl/pull/361

magnars:
Excellent 👍👍

vemv:
Thank you both for the understanding! 🍻

